PR #13, opened 2020-09-07
mentioned podAnnotations in values.yaml

ThomasSteinbach:
...as this variable will be used here https://github.com/nextcloud/helm/blob/master/charts/nextcloud/templates/deployment.yaml#L30


tvories (dismissed):
Looks good, but can you [sign off on the DCO](https://github.com/nextcloud/helm/pull/13/checks?check_run_id=1082973793) and bump the chart version?  At this time that would be 2.5.8

PR #18, opened 2020-09-24
Fix matchLabels to include metrics

dgiebert:
Found this while enabling metric scraping on Rancher as this will otherwise also match the nextcloud deployment


billimek (approved):
LGTM

dgiebert:
This breaks already installed metrics as this field is immutable. How to add that to a README somewhere ?

PR #19, opened 2020-09-26
Upgrade to latest version and fix cron

pfuhrmann:
- Upgrading app to latest 19.0.3
- Upgrading dependencies versions
- Cron to run every 5 minutes as specified in the [Nextcloud docs](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/background_jobs_configuration.html#background-jobs)
- Partially revert  #17 as there seem to be an issue with `trusted_domains` when trying to call `cron.php` without ingress
```
root@nextcloud-69c99f6d76-x45qc:/var/www/html# curl -I --fail -L http://nextcloud:8080/cron.php
curl: (22) The requested URL returned error: 400 Bad Request
```

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 2.1.2
-appVersion: 17.0.0
+version: 2.2.0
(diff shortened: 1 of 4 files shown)


billimek (approved):
LGTM

PR #23, opened 2020-10-05
[Bugfix] Add component label to application-related resources

### LINKED ISSUE #22: HPA non-functional due to incorrect LabelSelector
# Issue

Nextcloud App pod HPA does not work

# Cause

The HPA is unable to get the CPU requested/usage/limit for the application pods.
image

# Why

The HPA references the `nextcloud` deployment. This deployment does not define a component label. Therefore, if you also deploy the `nextcloud-metrics` deployment, one of the following will happen:

1. Metrics are returned for both the metrics pod and application pod, and are averaged together (if resource requests/limits defined on metrics pod)
2. If no resource requests/limits are defined on the metrics pod, the metrics are returned, but the HPA throws an error, because while the application pod may have resources defined, the metrics pod does not, and the HPA is unable to compute and average requests/limits as a result

# What happens

image

The metrics server logs do not return an error, but their debug output is helpful in highlighting the issue further. In the above, the HPAs I have deployed for nginx ingress controllers specifically call out the component label in their request. These HPAs are working without issue.

# Suggested Fix

Add `app.kubernetes.io/component: app` to the nextcloud application deployment, and update the spec accordingly:
```yaml
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: nextcloud
      app.kubernetes.io/name: nextcloud
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: nextcloud
        app.kubernetes.io/name: nextcloud
        nextcloud-redis-client: 'true'
```

I am happy to open a PR for this issue if this seems like the correct approach.

WojoInc:
Signed-off-by: TJ Wesolowski <wojoinc@pm.me>

#### Special notes for your reviewer:
Closes #22 
This should be bugfix only, so did not bump minor version.
To test on existing deployed release, delete the `nextcloud` deployment and run a `helm upgrade`.

#### Checklist
- [x] Chart Version bumped
- [x] Fix Validated

#### Proof
Confirmed this fix addresses the issue.
image

image

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.2.0
+version: 2.2.1
 appVersion: 19.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -7,6 +7,7 @@ metadata:
     helm.sh/chart: {{ include "nextcloud.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/component: app
   {{- if .Values.deploymentAnnotations }}
   annotations:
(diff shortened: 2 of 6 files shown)


billimek (approved):
LGTM

Issue #26, opened 2020-10-14
Add startupProbe for container


lober-io:
Due to the installation in the initial setup, the startup takes some time. Currently, this is caught through (high) initialDelaySeconds values.  A better solution would be to implement startupProbe as describe [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes). 

If you agree, I would create a pr with this change.

billimek:
@lober-io this is a good idea, and `startupProbe` certainly makes sense for this use-case.

`startupProbe` was introduced in Kubernetes starting with v1.16.  It may cause compatibility issues if we add this change.  Personally, I believe we should add this without a version check.

lober-io:
@billimek great, I will start working on this :+1:   

If we use the same pattern as in the other checks (enabled: true/ false), this would be no problem with the older k8s versions. The user just needs to set enabled: false, or? Then should be a hint in the readme sufficient.

billimek:
@lober-io a conditional, maybe with the enabled defaulting to `false` would be a great way to go about this IMO.

webwurst:
Can this be closed?

Timoses:
Why is this not enabled by default? I just spent 2 hours looking for the reason why the pod "crashed" during `rsync` ; ).

Btw: `startupProbe` was enabled as default since 1.18 (https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).

jessebot:
@tvories or @provokateurin what do you think of enabling this by default? A few people have brought this up, and it seems like it could be helpful. If anyone in this thread would like to submit a PR, I can review it provided everyone thinks this is a good idea. If k8s enables it by default, it is probably safe these days. A bit of documentation on the different probes would also probably be helpful to include in our README.

jessebot:
As this issue was originally for adding a startupProbe and it was done in #27, and we have docs on it here:
https://github.com/nextcloud/helm/tree/main/charts/nextcloud#configuration

And they are all present here:
https://github.com/nextcloud/helm/blob/bf6cc4a9df0b3bffd3915dc940ddbec71976429e/charts/nextcloud/values.yaml#L467-L490

[...]

PR #27, opened 2020-10-17
Add StartupProbe

lober-io:
As discussed in #26  I added the startupProbe (default: false) and adjusted the liveness and readiness probes.


billimek (approved):
LGTM

PR #50, opened 2020-12-14
Update values.yaml - Fix cert-manager apiVersion

dharmendrakariya:


tvories:
Nice catch.  You will need to [bump the chart version](https://github.com/nextcloud/helm/blob/0913ce2f362dbd967cb553146191314f57b9eac9/charts/nextcloud/Chart.yaml#L3) and [sign the DCO](https://github.com/nextcloud/helm/pull/50/checks?check_run_id=1549636415)

dharmendrakariya:
I dont know why but DCO thing is not happening, I tried 2 -3 times, what should I do ?

tvories:
@dharmendrakariya you pushed after ammending the DCO commit?  Also, we will need to bump the chart version to 2.5.4 since we just pushed 2.5.3

dharmendrakariya:
> @dharmendrakariya you pushed after ammending the DCO commit? Also, we will need to bump the chart version to 2.5.4 since we just pushed 2.5.3

Yeah, I think so, will do that tomorrow ✌️

dharmendrakariya:
@tvories hey brother, done, please have a look.

tvories:
Merged!  Thanks @dharmendrakariya

PR #54, opened 2020-12-22
Set redis default password and ENV REDIS_HOST_PASSWORD

tvories:
The nextcloud deployment needs the ENV variable `REDIS_HOST_PASSWORD` set in order for authenticated redis to not have issues.  This should fix #29

See https://github.com/nextcloud/docker/pull/1232 for additional details

PR #56, opened 2020-12-31
Added redis password to config

healthsyscn:
Following up #54 to add the redis password to the redis.config.php file

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.5.0
+version: 2.5.3
 appVersion: 19.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


tvories:
@billimek this definitely appears needed to complete the fix in #54

billimek:
Hi Hi @healthsyscn can you please [sign the DCO](https://github.com/nextcloud/helm/pull/56/checks?check_run_id=1631639365) and bump the chart version in the `Chart.yaml` file in order to signify a change in the chart?  

At the time of this writing, `2.5.3` is probably a good version to use.

tvories:
@healthsyscn I already modified the chart version number if you can sign the DCO.

ishioni:
Guys, i just checked a fresh install (with patches included), and it seems that nextcloud refuses to work with redis 6 and an empty password. I'm getting 303 replies when logging in and absolutely no log output about the problem. Can someone confirm it SHOULD work? Otherwise we should set some default password

tvories:
@ishioni I would agree that we should default to a password even if none is provided to resolve the redis 6 password woes.  I'll see if I can get something pushed here that fixes that.

tvories:
I'm going to approve this and we can track the "setting a default password" conversation in another issue.

tvories:
This has been addressed in #89 along with some other improvements.  Closing in favor of that PR.

PR #72, opened 2021-02-09
document hpa dependency

marvinosswald:
Added some info on clustering and its dependency on a shared persistence.

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -283,7 +283,7 @@ nextcloud:
 ```
 
 ## HPA (Clustering)
-If you want to have multiple nextcloud container regardless if dynamic sized or static - you need to use a shared persistence between the containers.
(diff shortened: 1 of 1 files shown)


charts/nextcloud/README.md
```diff
@@ -281,3 +281,13 @@ nextcloud:
       # note that Kubernetes currently requires cpu or memory requests and limits before hugepages are allowed.
       memory: 1Gi
 ```
+
+## HPA (Clustering)
+If you want to have multiple nextcloud container regardless if dynamic sized or static - you need to use a shared persistence between the containers.
```
tvories, line 286:
```suggestion
If you want to have multiple Nextcloud containers, regardless of dynamic or static sizes, you need to use shared persistence between the containers.
```

tvories (changes requested):
Updated wording a bit

tvories:
@marvinosswald you need to sign the DCO on your last commit and we can merge.

PR #86, opened 2021-02-20
Fix metrics secretKeyRef

timtorChen:
Currently, the metrics refer only to default secret name. This cause the metric pod cannot find the secret reference, if you use the different chart value of `.nextcloud.existingSecret.secretName`.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.5.6
+version: 2.5.7
 appVersion: 19.0.3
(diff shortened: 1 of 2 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.5.5
+version: 2.5.7
```
tvories, line 3: Can you bump this to 2.5.12?

tvories (changes requested):

charts/nextcloud/Chart.yaml
timtorChen: done

tvories (approved):

PR #98, opened 2021-03-11
Add securityContext and containerPort configuration

menardorama:
# Pull Request

## Description of the change

Hi,

This PR is intended to permit to run Nextcloud on a "no privileged" Kubernetes cluster. We are enforcing security by disabling container to run as root, for Nextcloud we need to rebuild the docker images in order to pass specific configuration in order to bind a non-privileged port.

This PR introduce the possibility to define securityContext (runAsNonRoot....) and change the containerPort

## Benefits

Run a Kubernetes cluster where PodSecurityPolicy (or something else like OPA) forbid to run as root.

## Possible drawbacks

From my point of view none.

## Applicable issues

- fixes #9 

## Additional information

None

## Checklist 

- [x] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] (optional) Variables are documented in the README.md

PR #115, opened 2021-05-20
Use service account name also for cronjob

aspettl:
# Pull Request

## Description of the change

When pod security policies (PSP) are used, then the securityContext typically has e.g. a "runAsUser" attribute and a (non-default) service account. The cronjob uses the same security context as the regular deployment, so it should also use the same service account. With PSP enabled, a pod created with the "wrong" service account will not be able to start.

## Benefits

The current chart works with PSP (when security context and service account are set correctly). But when you enable the cronjob, it will not start - this is fixed here.

## Possible drawbacks

Cronjob pod has more permissions than it needs. But this is already the case, it starts "curl" as root by default, which is just not needed. Fixing this would be a breaking change and should be considered in a separate PR.

## Applicable issues

- PSP support: https://github.com/nextcloud/helm/pull/105
- Non-root support: https://github.com/nextcloud/helm/pull/98

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.6.4
+version: 2.6.5
 appVersion: 19.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


tvories:
Thanks for the submission!

PR #134, opened 2021-08-05
Add Dependabot configuration from template repo

splitt3r:
Adds Dependabot updates for GitHub Actions.

See: https://github.com/nextcloud/.github/blob/master/.github/dependabot.yml

Example: https://github.com/splitt3r/helm/pull/2

After merging this you will be able to trigger manual checks at https://github.com/nextcloud/helm/network/updates


tvories:
Thanks @splitt3r !

Issue #58, opened 2021-01-04
Include path in ingress


krisko:
Hi, could you please include option to configure path in ingress.yaml?

```
spec:
  rules:
  - host: {{ .Values.nextcloud.host }}
    http:
      paths:
      - backend:
          serviceName: {{ template "nextcloud.fullname" . }}
          servicePort: {{ .Values.service.port }}
        path: {{ .Values.ingress.path }}
```

jessebot:
Thanks for opening the issue! This was added in https://github.com/nextcloud/helm/commit/7d840d368e17be1423359e841e50eacda802f717 so I will close this now. If you have further issues with ingress, please open another ticket and I'm happy to help look into it with you :)

marksaitis:
There is not a single working example of a nextcloud helm values ingress config to put nextcloud under domain.com/nextcloud . None of the conventional methods work

PR #139, opened 2021-08-14
Adding configurable ingress path

christopherklug:
according to https://github.com/nextcloud/helm/issues/58

# Pull Request

## Description of the change

na

## Benefits

Allows to deploy nextcloud on other ingress paths than root

## Possible drawbacks

None


tvories:
@christopherklug you'll need to bump the chart version for this to pass the lint test.  As of now, you could use `2.7.4`

christopherklug:
Did check the changes a second time and fixed introduced problems. Hope i didn't miss something, but the patch worked for me with and without an ingress path. In my opinion ready to merge.

charts/nextcloud/templates/cronjob.yaml
```diff
@@ -51,9 +51,9 @@ spec:
                 - "--fail"
                 - "-L"
               {{- if .Values.ingress.tls }}
-                - "https://{{ .Values.nextcloud.host }}/cron.php"
+                - "https://{{ .Values.nextcloud.host }}{{ .Values.ingress.path }}/cron.php"
               {{- else }}
-                - "http://{{ .Values.nextcloud.host }}/cron.php"
+                - "http://{{ .Values.nextcloud.host }}{{ .Values.ingress.path }}/cron.php"
```
mrueg, line 56: With the current default it renders to `http://host//cron.php` is that intended?

PR #147, opened 2021-09-01
Fix MYSQL_ env variables conditional

danielwegener:
# Pull Request

I want to set all POSTGRES_ values from a mounted secret (even POSTGRES_DB_FILE, POSTGRES_USER_FILE and POSTGRES_PASSWORD_FILE). Therefore I have set:
- internalDatabase.enabled: false
- externalDatabase.type: "" # in order to not have the environment variables set by the deployment.yaml

The `file_env` behaviour of /entrypoint.sh does not support both (POSTGRES_DB and POSTGRES_HOST_FILE), it will `echo >&2 "error: both $var and $fileVar are set (but are exclusive)"`. 

The unfortunate behaviour of the chart as of today is to "fall back" to MYSQL_ variables which in turn will trigger the "mysql" path in the entrypoint.sh. This does not really make sense.

Tried alternatives without this PR:
I can not override/unset POSTGRES_* via extraEnv, because they are already set. I can not override the entrypoint to unset them.

## Description of the change

This PR does not clearly introduce a new configuration capability but somewhat compensates a weird configuration quirk and as a side effect...

## Benefits

...  allows configuring database env variables freely.

## Possible drawbacks

Likely none, `mysql` is the default value for `externalDatabase.type` anyway

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.8.2
+version: 2.8.3
 appVersion: 21.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


danielwegener:
alternative approach would be to check for `{{- else if .Values.externalDatabase.enabled }}` in deployment.yaml:88

Dem0n3D:
Please, merge this.

Jeroen0494:
@danielwegener any change you could revisit this PR? I'd like this fix as well. For now, I solved it by putting the MySQL env variables directly under .Values.nextcloud.env and extraEnv.

Issue #119, opened 2021-06-21
Upgrade dependencies


cyrilcros:
Tied to #106 - please upgrade the Bitnami dependencies....

jessebot:
All of these dependencies have been upgraded since this issue was opened, so I will close it.

PR #152, opened 2021-10-07
Update mariadb

ariep:
# Pull Request

## Description of the change

Update mariadb dependency chart to most recent version.

## Benefits

Enjoy all benefits of new and improved mariadb, both the program itself as well as helm packaging.

Our primary motivation for doing this was that the newer chart version allows setting custom kubernetes labels on the mariadb statefulset and its pods, which is something we need for our backup/restore solution that uses velero.

## Possible drawbacks

1. The mariadb chart has made some incompatible changes, see [their upgrade notes](https://github.com/bitnami/charts/tree/master/bitnami/mariadb/#to-800). Depending on a user's setup (replicated or standalone) this may require manual intervention when upgrading.
2. Because the mariadb chart has started using the bitnami "common library chart", it became harder to recreate the database's kubernetes service name, which is calculated by a helm template function. Nextcloud needs to know this service name because it uses that as hostname to contact the database service. Luckily there is an [elegant principled solution in helm](https://github.com/helm/helm/pull/9957), to allow calling a subchart's template functions, instead of having to copy their definition. The downside is that solution is only supported by a very recent helm version (3.7.0, released three weeks ago), so this would put rather strict helm version requirements on anyone using this helm chart.

There is no upgrade guide in the README currently, but 1. should really be documented somewhere.

If deemed necessary, maybe we could mitigate 2. by checking for the presence of `.Subcharts`, and defaulting to the old template function to compute the database service name.

## Applicable issues

- fixes a part of #119

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.8.3
+version: 2.9.0
 appVersion: 21.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:


varac:
@tvories Any chance to get this reviewed soon ?

varac:
@tvories  Awesome, thx !

PR #159, opened 2021-10-30
GKE ingress

### LINKED ISSUE #158: GCE Ingress only supports "ImplementationSpecific" as path type
# Problem

When using a [GCE/GKE Ingress](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress) on GCP and using values like these:

```
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.global-static-ip-name: nextcloud
    kubernetes.io/ingress.class: gce
    networking.gke.io/managed-certificates: nextcloud
    networking.gke.io/v1beta1.FrontendConfig: nextcloud-frontend-config
```
the deployment of the ingress fails with this error:

```
Translation failed: invalid ingress spec: only "ImplementationSpecific" path type is supported
```

# Possible Solutions

- :3rd_place_medal: Setting `ingress.enabled: false` and provision the ingress on your own. But this makes the `CronJob` fail because of "untrusted domain" issues as the CronJob then calls the service directly (https://github.com/nextcloud/helm/blob/master/charts/nextcloud/templates/cronjob.yaml#L59). So, you have to switch to AJAX cronjobs. That's my workaround right now.
- :1st_place_medal: Somehow magically detect that a GCE ingress is going to be used and set `pathType` to `ImplementationSpecific`.
- :1st_place_medal: Provide an option for the Helm values to specify the desired `pathType`.

zumbrunnen:
# Pull Request

## Description of the change

Ability to configure for use with a GKE Ingress (on Google Cloud Platform). Before, this was not possible (see #158).

## Benefits

These values can be specified to make Nextcloud work on GKE (with a GKE Ingress):

```yaml
ingress:
  enabled: true
  annotations:
    [...]
    kubernetes.io/ingress.class: gce
    [...]
  path: /*
  pathType: ImplementationSpecific
```

## Possible drawbacks

None. Other users don't have to change anything. Their setup is unaffected by this change.

## Applicable issues

- fixes #158

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.9.0
+version: 2.9.1
 appVersion: 21.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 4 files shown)


tvories:
Nice feature!  Thanks for the submission.

Issue #110, opened 2021-04-19
[Feature] Allow Different PVCs for Subpaths


Blackclaws:
In our use case we have a slow high-redundant and a fast low-redundant storage system. We would like to put the main application components into the fast storage but keep all data in the high-redundant storage. This is currently not possible.

Would you accept pull requests that add this functionality?

jessebot:
Thank you for bringing it up :) This has been fixed in #161, so I will close this ticket.

PR #161, opened 2021-11-02
add posibility to have data on a second pvc that can have independent limits, storage class from main pvc

xMAC94x:
# Pull Request

## Description of the change

implements partial #110.
We add the possibility to move data directory in a different PVC by setting a config flag.
This is DISABLED by default

## Benefits

In order to profit from different storage classes (e.g. high speed SSD and slow HDD) and special limits on the data.

## Possible drawbacks

If enabled creating a second PVC needs a few more ressources.
We might have problems if we want to delete this feature in the future with backwards compability.

## Applicable issues

no known

## Checklist 

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -338,6 +338,17 @@ persistence:
   accessMode: ReadWriteOnce
   size: 8Gi
 
+  ## Use an additional pvc for the data directory rather than a subpath of the default PVC
+  ## Useful to store data on a different storageClass (e.g. on slower disks)
+  nextcloudData:
+    enabled: false
+    subPath:
+    annotations: {}
+    # storageClass: "-"
+    # existingClaim:
+    accessMode: ReadWriteOnce
+    size: 8Gi
+
 resources: {}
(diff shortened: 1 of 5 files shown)


charts/nextcloud/values.yaml
```diff
@@ -330,6 +330,15 @@ persistence:
   ##
   # storageClass: "-"
 
+  ## Use a additional pvc for the data directory rather than a subpath of the default pvc
+  ## Useful to store data on a different storageClass (e.g. on slower disks)
+  nextcloudData:
+    enabled: false
+    subPath:
+    # storageClass: "-"
+    accessMode: ReadWriteOnce
```
tvories, line 339: Can we add the `existingClaim` example here?

```yaml
# existingClaim: "-"
```

xMAC94x: Hi @tvories  added the same existingClaim here and moved the whole section to the bottom of `persistence`.
Is it okay this way or should i also include the ` "-"` ?

tvories: No, this is great!  Thanks!

PR #219, opened 2022-04-03
Configurable HostAlias for cronjob

simoncaron:
# Pull Request

## Description of the change

The k8s cronjob uses the Nextcloud host when it's doing it's curl. In some cases, the host isn't resolvable inside the cluster so the cronjob end with a _Could not resolve host_ error.

## Benefits

Configurable hostaliases allows to define a host entry for the nextcloud host in the cronjob host file so it's able to curl properly the cron endpoint from inside the k8s cluster.

## Possible drawbacks

None

## Applicable issues

None

## Additional information

None

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.13.2
+version: 2.14.0
 appVersion: 23.0.2
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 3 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.14.1
+version: 2.15.0
```
jessebot, line 3: This needs to be the latest version

charts/nextcloud/values.yaml
```diff
@@ -312,6 +312,8 @@ cronjob:
     #  cpu: 100m
     #  memory: 128Mi
 
+  # hostAliases: {}
```
jessebot, line 315: This would also need to be updated in the configuration parameters section of the `README.md`

provokateurin:
This is not applicable anymore since the chart doesn't use the k8s cronjob anymore.

Issue #240, opened 2022-06-07
Your web server is not properly set up to resolve "/.well-known/webfinger".


icebrian:
Nextcloud reports my rewrite are not setup correctly with message:

Your web server is not properly set up to resolve "/.well-known/webfinger".

I am using 24.0.1-apache image together with an nginx ingress controller, thus using provided configuration both in .htaccess file (part of container) and ingress controller configuration which is:

```
      server_tokens off;
      proxy_hide_header X-Powered-By;

      client_max_body_size 16G;
      fastcgi_read_timeout 3600s;
      proxy_buffering off;
      gzip off;

      rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
      rewrite ^/.well-known/host-meta /public.php?service=host-meta last;
      rewrite ^/.well-known/host-meta.json /public.php?service=host-meta-json;
      location = /.well-known/carddav {
        return 301 $scheme://$host/remote.php/dav;
      }
      location = /.well-known/caldav {
        return 301 $scheme://$host/remote.php/dav;
      }
      location = /robots.txt {
        allow all;
        log_not_found off;
        access_log off;
      }
      location ~ ^/(?:build|tests|config|lib|3rdparty|templates|data)/ {
        deny all;
      }
      location ~ ^/(?:autotest|occ|issue|indie|db_|console) {
        deny all;
      }
```

Northcode:
the line for "webfinger" is supposed to say:
```
rewrite ^/.well-known/webfinger /index.php/.well-known/webfinger last;
rewrite ^/.well-known/nodeinfo /index.php/.well-known/nodeinfo last;
```
instead of 
```
rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
```

I made a PR to update the example in values.yaml.

PR #241, opened 2022-06-07
update ingress snippet comment in values.yaml

Northcode:
# Pull Request

## Description of the change

Update well-known webfinger and nodeinfo redirects to match documentation in https://docs.nextcloud.com/server/23/admin_manual/issues/general_troubleshooting.html#service-discovery

## Benefits

People just uncommenting the snippet will now have correct webfinger config, and no warnings in the nextcloud settings.

## Possible drawbacks

None that I can see

## Applicable issues

#240 

## Additional information

...

## Checklist

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.8
+version: 3.5.9
 appVersion: 26.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.7
+version: 3.5.8
 appVersion: 26.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -30,8 +30,8 @@ ingress:
   #  nginx.ingress.kubernetes.io/server-snippet: |-
   #    server_tokens off;
   #    proxy_hide_header X-Powered-By;
-
-  #    rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
+  #    rewrite ^/.well-known/webfinger /index.php/.well-known/webfinger last;
+  #    rewrite ^/.well-known/nodeinfo /index.php/.well-known/nodeinfo last;
   #    rewrite ^/.well-known/host-meta /public.php?service=host-meta last;
   #    rewrite ^/.well-known/host-meta.json /public.php?service=host-meta-json;
   #    location = /.well-known/carddav {


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 2.14.4
+version: 2.14.5
```
jessebot, line 3: Please update this chart version to one above the default branch.

charts/nextcloud/values.yaml
```diff
@@ -30,7 +30,9 @@ ingress:
   #    server_tokens off;
   #    proxy_hide_header X-Powered-By;
 
-  #    rewrite ^/.well-known/webfinger /public.php?service=webfinger last;
+  #    rewrite ^/.well-known/webfinger /index.php/.well-known/webfinger last;
+  #    rewrite ^/.well-known/nodeinfo /index.php/.well-known/nodeinfo last;
+
```
jessebot, line 35: Please fix the trailing spacing here

jessebot (changes requested):

Jeroen0494:
@Northcode any change you could revisit this PR? I can confirm this fix works.

Northcode:
@Jeroen0494 
Hi. Sorry for the late reply. I had forgotten about this.
I've fixed the things jessebot was complaining about and rebased on upstream.
Not sure how to proceed as github still says jessebot is requesting changes.

Northcode:
@tvories 
I think it was complaining that there was trailing whitespace on line 35 of the [old commit](https://github.com/nextcloud/helm/commit/c3aaa3d8b7bffdc6259060c498e71c36c24782b4).
I've corrected that in my [updated code](https://github.com/nextcloud/helm/commit/7e209a1a453bd4faee4a6d33f9baf1d7d3dc3330), but it doesn't look like jessebot has rerun, its still complaining about the old commit.
I marked those corrections are resolved, is there something else I have to do for it to detect my changes?

Jeroen0494:
> I marked those corrections are resolved, is there something else I have to do for it to detect my changes?

You can go to the "1 change requested" option, click the menu and select "Request re-review" or something.
image

charts/nextcloud/values.yaml
```diff
@@ -31,7 +31,9 @@ ingress:
...
```
tvories, line 36: Is there any reason we need this newline here?  It makes sense to group the rewrites together.

jessebot: yeah, let's keep the rewrites together, and maybe also remove the space on line 33? This way, it's clean that the entire block of commented stuff in `nginx.ingress.kubernetes.io/server-snippet` is related.

Northcode:
@jessebot oh I see. In that case sorry for assuming your uhm.. type of consciousness? I guess? haha

Anyways I've removed the whitespace now on both line 36 and 33 and rebased onto upstream again, incrementing the chart version to 3.5.8

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.7
+version: 3.5.8
```
```suggestion
version: 3.5.9
```

jessebot (changes requested):
chart.yaml needs bumping. Left the suggestion so you can just click the button and we can merge this. Thank you for fixing the whitespace :)

charts/nextcloud/Chart.yaml
jessebot: actually, it looks like I can just do the commit for you and then we'll just be coauthors :)

jessebot (approved):

PR #249, opened 2022-06-28
Fix: Bump to 24.0.2

mrueg:
# Pull Request

## Description of the change

Forgot to bump the tag in the previous commit, thanks to @alesc for mentioning it

## Applicable issues

- fixes #

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.lock
+++ b/charts/nextcloud/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 11.6.8
+  version: 11.6.10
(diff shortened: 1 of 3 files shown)


mrueg:
@tvories  
Should we add something to _helpers.tpl that will take the value from appVersion if the tag is not set? 
That way we only need to update in one place.
Similar to https://github.com/prometheus-community/helm-charts/blob/47d3b08e980cd0862e28f7d7f49c07dd7b9b7091/charts/prometheus-node-exporter/templates/_helpers.tpl#L61

tvories:
Yes, was thinking that earlier. I actually think that's why I missed it was because I thought we already had it.

PR #253, opened 2022-07-04
Use .Chart.AppVersion for image tag

mrueg:
# Pull Request

## Description of the change

@tvories  this uses AppVersion if the tag key is unset and the tag key if set. I also added a flavor key which allows users to switch to a different image type (e.g. nginx or something else).

## Checklist 


tvories:
Looks good, thanks!

PR #262, opened 2022-07-30
Registry for postgresql initContainer

fdelucchijr:
Signed-off-by: Fernando De Lucchi <69133647+fdelucchijr@users.noreply.github.com>

# Pull Request

## Description of the change

Just parses registry from values.yaml Postgres image key

### DIFF
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -409,7 +409,7 @@ spec:
       {{- else if .Values.postgresql.enabled }}
       initContainers:
       - name: postgresql-isready
-        image: {{ .Values.postgresql.image.registry }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
+        image: {{ default "docker.io" .Values.postgresql.image.registry }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
         env:
           - name: POSTGRES_USER
             valueFrom:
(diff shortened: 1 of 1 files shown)


charts/nextcloud/templates/deployment.yaml
```diff
@@ -409,7 +409,7 @@ spec:
       {{- else if .Values.postgresql.enabled }}
       initContainers:
       - name: postgresql-isready
-        image: {{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
+        image: {{ .Values.postgresql.image.registry }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
```
MichaelSp, line 412: why not wrote `repository: docker.io/image` instead of 

```yaml
registry: docker.io
repository: image
```

The proposed change would't work for the case where you dont want to provide a registry. It would actually be a breaking change. 
In that case it would make sense to provide a default:
```suggestion
        image: {{ default "docker.io" .Values.postgresql.image.registry }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
```

MichaelSp (changes requested):

MichaelSp:
btw: @fdelucchijr  I'm not a maintainer here

jessebot (changes requested):
You need to bump the [helm chart version](https://github.com/nextcloud/helm/blob/master/charts/nextcloud/Chart.yaml#L3) when you make a change as per the checklist in the [pull request template](https://github.com/nextcloud/helm/blob/master/.github/pull_request_template.md#checklist). You should also update the README.md and values.yaml with any new parameters you are introducing.

PR #117, opened 2021-06-08
Use securityContext for deployment and cronjob

nlaille:
# Pull Request

## Description of the change

Do a separation of securityContext for deployment and cronjob.

## Benefits

Possibility to set `fsGroup` and `fsGroupChangePolicy` on deployment. 
With the current setup, `fsGroup` and `fsGroupChangePolicy` cannot be set because cronjob and deployment uses the same securityContext and cronjob securityContext doesn't have `fsGroup` field.

## Possible drawbacks

This PR should backward compatible.

## Applicable issues

- fixes #

## Checklist 

- [ ] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] (optional) Variables are documented in the README.md

PR #269, opened 2022-08-16
Add security context for the cronjob, main container, main pod, and ngnix container

### LINKED ISSUE #186: Possible wrong extraSecurityContext
In `values.yaml` you have `nextcloud.extraSecurityContext` but it's not used anywhere in the helm chart.

https://github.com/nextcloud/helm/blob/master/charts/nextcloud/values.yaml#L174

What is used is `.securityContext`

https://github.com/nextcloud/helm/blob/master/charts/nextcloud/templates/deployment.yaml#L445

### LINKED ISSUE #252: securityContext is used both on pod and container level
The securityContext settings are used on a pod level in the `deployment.yaml` and on container level in the `cronjob.yaml`. Unfortunately those secuirtyContexts are not compatible. I have to set `securityContext.fsGroupChangePolicy` for the nextcloud deployment, but that setting does not exist on container level and therefore the Helm deployment fails.

Maybe it would make sense to split the security context for the nextcloud deployment and the cronjob?

Timmy1e:
# Pull Request

## Description of the change

Allows users to set the securityContext for the NextCloud pod, NextCloud container, nginx container, and the cronjob.

Based on #117 by @nlaille.

## Benefits

Allows users to set the securityContext for the NextCloud pod, NextCloud container, nginx container, and the cronjob.

## Possible drawbacks

Users can overwrite the default `runAs` property.

## Applicable issues

- fixes #186
- fixes #252

## Checklist 

### DIFF
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -121,10 +121,16 @@ spec:
           successThreshold: {{ .Values.startupProbe.successThreshold }}
           failureThreshold: {{ .Values.startupProbe.failureThreshold }}
         {{- end }}
-        volumeMounts:
-{{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if .Values.nextcloud.securityContext}}
+        securityContext:
+          {{- with .Values.nextcloud.securityContext }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
+        {{- end }}
+        volumeMounts:
+        {{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
       {{- if .Values.cronjob.enabled }}
       - name: {{ .Chart.Name }}-cron
         image: {{ include "nextcloud.image" . }}
@@ -150,6 +156,12 @@ spec:
 {{- include "nextcloud.env" . | indent 8 }}
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if .Values.cronjob.securityContext}}
+        securityContext:
+          {{- with .Values.cronjob.securityContext }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
+        {{- end }}
         volumeMounts:
 {{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
       {{- end }}
@@ -192,6 +204,12 @@ spec:
 
         resources:
 {{ toYaml .Values.nginx.resources | indent 10 }}
+        {{- if .Values.nginx.securityContext}}
+        securityContext:
+          {{- with .Values.nginx.securityContext }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
+        {{- end }}
         volumeMounts:
         - name: nextcloud-main
           mountPath: /var/www/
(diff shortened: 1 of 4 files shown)


Timmy1e:
@jessebot & @tvories, I've updated the PR, it's ready updated and for review.

provokateurin:
Sorry I forgot to review this yesterday. I'll try to review it tomorrow if I don't forget it again :sweat_smile:

charts/nextcloud/templates/deployment.yaml
```diff
@@ -121,10 +121,16 @@ spec:
...
         resources:
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if .Values.nextcloud.securityContext}}
+        securityContext:
+          {{- with .Values.nextcloud.securityContext }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
+        {{- end }}
+        volumeMounts:
+        {{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
```
provokateurin, line 133: Could you move this back above? I don't think there's a reason for this unrelated change

provokateurin, line 133: Could you move this back above? I don't think there's a reason for this unrelated change

charts/nextcloud/templates/deployment.yaml
```diff
@@ -147,11 +153,17 @@ spec:
         {{-   end }}
         {{- end }}
         env:
-{{- include "nextcloud.env" . | indent 8 }}
+        {{- include "nextcloud.env" . | indent 8 }}
```
provokateurin, line 156: Also unrelated change

charts/nextcloud/templates/deployment.yaml
```diff
@@ -147,11 +153,17 @@ spec:
...
 {{ toYaml .Values.resources | indent 10 }}
+        {{- if .Values.cronjob.securityContext}}
+        securityContext:
+          {{- with .Values.cronjob.securityContext }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
+        {{- end }}
         volumeMounts:
-{{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
+        {{- include "nextcloud.volumeMounts" . | trim | nindent 8 }}
```
provokateurin, line 166: Also unrelated change

charts/nextcloud/templates/deployment.yaml
```diff
@@ -308,23 +326,23 @@ spec:
...
+        # Will mount configuration files as www-data (id: 33) for nextcloud
         fsGroup: 33
-        {{- if .Values.securityContext }}
-        {{- with .Values.securityContext }}
-        {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- if .Values.securityContext }}
+          {{- with .Values.securityContext }}
+            {{- toYaml . | nindent 8 }}
+          {{- end }}
```
provokateurin, line 345: Since there is `.Values.nextcloud.podSecurityContext` now and `.Values.securityContext` is not documented anywhere, maybe just remove it?

provokateurin (changes requested):

charts/nextcloud/templates/deployment.yaml
Timmy1e: I moved it so all containers in this pod have `resources`, `securityContext`, `volumeMounts` in the same order.

charts/nextcloud/templates/deployment.yaml
Timmy1e: I can remove it, but the reason I kept it in was for some backwards compatibility.

charts/nextcloud/templates/deployment.yaml
provokateurin: Alright :)

charts/nextcloud/templates/deployment.yaml
provokateurin: Yeah, but I wonder how many people actually rely on it since it's undocumented. Could you instead add a deprecation comment?

Timmy1e: I've added the deprecation comment to the `values.yaml` file.

provokateurin:
Will review tomorrow :) I only work two days a week, so there is always a delay

provokateurin (approved):

Timmy1e:
@jessebot Your assumption about `podSecurityContext` is correct, and it does appear that I have forgotten to update the description for `nextcloud.podSecurityContext`. Good catch! Feel free to open a PR if to fix it.

provokateurin:
Run `helm template` to see the rendered output. This will probably point you to the source of the error.

PR #274, opened 2022-08-31
Fix issue with service selector when running multiple instances

mikaelh:
Signed-off-by: Mikael Hammarin <mhammarin@ecit.com>

# Pull Request

## Description of the change

This fix make it possible to run multiple helm instances in the same namespace without traffic accidently being sent to all of them (service selector is not unique per instance). 

## Benefits

This allow multiple parallel instances of Nextcloud to be run in a single namespace without colliding.

## Possible drawbacks

This change make it impossible to lunch multiple deployments which will receive traffic from the same ingress.
However, I can't think of a reason why it would be intended behavior for multiple deployment to use the same service selector.

## Applicable issues

NA

## Additional information

NA

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.2.1
+version: 3.2.2
 appVersion: 24.0.5
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


mikaelh:
ping @tvories :)

mikaelh:
any update? this is a small fix and non-breaking change  @tvories

mikaelh:
@tvories bumped version number

PR #278, opened 2022-09-06
Fix(nginx-config): Listen port based on nextcloud.containerPort

Nold360:
Signed-off-by: nold <nold@gnu.one>

# Pull Request

## Description of the change

If `nextcloud.containerPort` is set, it will configure nginx to listen to the specified port.

## Benefits

Allows nginx container to run rootless & follows the expected behavior of `nextcloud.containerPort`.

## Possible drawbacks

Unexpected port change on update?!

## Applicable issues

- fixes #

## Checklist 

### DIFF
--- a/charts/nextcloud/templates/nginx-config.yaml
+++ b/charts/nextcloud/templates/nginx-config.yaml
@@ -44,7 +44,7 @@ data:
         }
 
         server {
-            listen 80;
+            listen {{ .Values.nextcloud.containerPort | default 80 }};
 
             # HSTS settings
             # WARNING: Only add the preload option once you read about
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.1.1
+version: 3.1.2
 appVersion: 24.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:


tvories:
Thanks for the submission @Nold360 .  can you bump the chart version?

Nold360:
> Thanks for the submission @Nold360 . can you bump the chart version?

fixed

relrod:
+1, this seems necessary for OpenShift.

For now I've worked around this using the normal apache image, as an image stream and layered a change on top of it to make it listen on another port, but this is a much cleaner solution.

I also think the readiness probes also need to be able to have their ports configured, because that was the next issue I ran into on OpenShift. They seem to hardcode `port: http` right now.

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -44,7 +44,7 @@ data:
         }
 
         server {
-            listen 80;
+            listen {{ .Values.nextcloud.containerPort | default 80 }};
```
provokateurin, line 47: Could you move the default value into the values file? That would make it easier to reuse for https://github.com/nextcloud/helm/pull/316

provokateurin (changes requested):

charts/nextcloud/templates/nginx-config.yaml
jessebot: Sorry, there's 3 PRs for this that have been open for a couple of months (except mine that I opened today). @provokateurin, are you asking that that `nextcloud.containerPort` in values.yaml be set to `80`?

@Nold360 if you don't have time, I can get this done via github suggestion commits.

provokateurin: Yes

jessebot: I couldn't figure out how to do a suggest on a new file, so I'm modifying #384 to include this change. Will close my PR if the other PRs are moved forward first.

PR #237, opened 2022-05-23
bump to nextcloud 24.0.1

varet80:
# Pull Request

## Description of the change

Create an ugprade release for Next cloud24.0.1

## Checklist 

- [ ] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] (optional) Variables are documented in the README.md

PR #280, opened 2022-09-11
bump to 24.0.5

MichaelSp:
supersedes #237

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 3.1.1
-appVersion: 24.0.4
+version: 3.1.2
+appVersion: 24.0.5
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
 - nextcloud


lion24:
Hi there,

Couldn't this process automated ? 
That's really painful to wait for a maintainer to approve trivial changes like this. We could miss some patch like the one in #237 

I'm sure we can leverage gh workflows for this.

Best regards,
Lionel.

MichaelSp:
ping @tvories

tvories:
@lion24 it should be automated, yes.  Maybe not auto-approved (unless it's a minor update), but at the very least auto-opened.  I think it would sometimes be nice to batch a few different PRs into a single release, but I don't have the time to rewrite some of the release workflows.

lion24:
@tvories thanks for the answers.

Nextcloud upgrade can be picky because before upgrading to the next major we need to ensure we have the latest point release as explained here: https://docs.nextcloud.com/server/latest/admin_manual/maintenance/upgrade.html otherwise I guess this can be dangerous as I guess their migration scripts assume you have the latest previous release installed, if not we can have some corruption or worse a broken installation.

> I think it would sometimes be nice to batch a few different PRs into a single release

Yes, that would work, but as said above, we need to ensure that we don't bump to a major nextcloud release before having installed the latest point release.

I don't have a lot of experience with gh actions, but I can have a look.

Cheers.
Lionel.

MichaelSp:
In this repo you can always bump to the latest version. No matter what. In the end you don't known what version (of this helm-chart) people have installed. If you are truly concerned about it, you should break the upgrade process and stay in maintenance mode until the admin resolves it. You cannot really roll-back the helm-chart from within the pod.

However, if you automated it, it would be nice to bump the same version-part in the chart as well (major increase in NC bumps major in this chart, etc)

PR #282, opened 2022-09-12
adds deploymentLabels parameter to values.yaml

tmm1234:
deploymentLabels is referenced in lines 11-13 of the deployment template but is not present in the values.yaml file

Signed-off-by: tmm1234 <75389007+tmm1234@users.noreply.github.com>

# Pull Request

## Benefits

supports the expected workflow of a helm chart and improve readability

## Possible drawbacks

/

## Applicable issues

/

## Checklist 


charts/nextcloud/values.yaml
```diff
@@ -12,6 +12,7 @@ nameOverride: ""
 fullnameOverride: ""
 podAnnotations: {}
 deploymentAnnotations: {}
+deplomentLabels:
```
jessebot, line 15: There is a typo in the word `deployment` here. Could you please fix that?

jessebot (changes requested):
Can you also please update the version for Chart.yaml? Other than that, it looks good and thanks for catching this :)

charts/nextcloud/values.yaml
tmm1234: done :)

jessebot (approved):

PR #292, opened 2022-10-14
Use cron instead of webcron

### LINKED ISSUE #55: System cron instead of webcron?
Apps like News currently [require system cron](https://github.com/nextcloud/news/issues/1005), which by my understanding isn't available with this helm chart by default.

Is it possible to run `cron.sh` as in some of the docker compose setups, instead of webcron?

### LINKED ISSUE #215: cronjobs no longer running
per this issue webcron isnt even supported but its in the docs and in this helm chart
https://github.com/nextcloud/server/issues/30899#issuecomment-1025430197

i have a warning in my ui that jobs have not run for days after upgrading to 23.0.3 using this official chart. using ajax never seemed to work for me and the system cron doesnt seem implemented in this container.

suggestions for running these jobs?

kretzlaff:
# Pull Request

## Description of the change

This change replaces webcron with system cron.
The implementation is similar to the official [nextcloud/docker/stack.yml](https://github.com/nextcloud/docker/blob/master/stack.yml#L29-L36)  configuration.

An additional container named {{ .Chart.Name }}-cron is created.
The command is hardwired to /cron.sh

To ensure the environment variables as well as the volumes match the main container the environment variables and volume mounts are moved to a template. The deployment imports the templates for both containers.

## Benefits

Uses the recommended method to execute background tasks.

## Possible drawbacks

- The webcron configuration is removed. Might introduce a breaking change when special handling is configured.
- Pod now contains an additional container (Similar to official docker-compose linked above)

## Applicable issues

- fixes #55 
- fixes #215 

## Checklist


jld3103:
Doesn't this break instances with more than one replica as per https://github.com/nextcloud/helm/issues/55#issuecomment-790805004 ?

kretzlaff:
I've seen this comment. To ensure it won't affect anything I currently run my setup with 2 replicas. So far it worked perfectly fine without any issues.

jld3103:
Right maybe it works, but it still doesn't seem right to me to run two cronjobs

kretzlaff:
I tried to dig a bit deeper.

According to cron.php there is a limit on how many jobs can run on a single instance.
See here https://github.com/nextcloud/server/blob/master/cron.php#L139-L142

Diving one level deeper to the jobList->getNext() method we can see that once a job is fetched it is immediately updated in the database to not be triggered twice. A second, third, ... nextcloud cron instance would just take the next one from the queue.
https://github.com/nextcloud/server/blob/master/lib/private/BackgroundJob/JobList.php#L208-L262

Since both these mechanisms exist it should not be possible to run the same job simultaneously in multiple instances.
However there will still be more cron jobs in parallel. BUT since you have such a huge nextcloud setup that you need to run multiple instances you're most likely happy that more background jobs are executed since they have to handle lots of data.
I have no idea how background jobs are generated but I'd assume the more a setup grows the more background jobs are generated which need to be run.

jld3103:
Then it's actually good I guess. Running more cronjobs on a big instance with multiple replicas makes sense :+1:

jld3103:
Thanks a lot for the PR, I've been maintaining my own version of the chart with exactly this sidecar cron image because I only run a single replica and knew it was fine, but now I can drop it!

remram44:
The README still says:

> This chart can utilize Kubernetes CronJob resource to execute background tasks.

Which is now totally wrong.

PR #316, opened 2022-11-27
Allow probe checks and Service to use containerPort if specified

relrod:
# Pull Request

## Description of the change

Note: This probably can and should fold into #278.

This allows `containerPort` to be used for the readiness/startup/liveness probes along with the `targetPort` of the Service, otherwise they default to 80 and fail if the httpd inside the container is listening on another port.

This is particularly important on OpenShift where by default you cannot access privileged ports.

## Benefits

Makes it easier for OpenShift users? :)

## Possible drawbacks

There should not be any - the old default is kept if `containerPort` isn't defined.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.3.3
+version: 3.3.4
 appVersion: 25.0.1
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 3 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.3.3
+version: 3.3.4
```
```suggestion
version: 3.6.1
```
Assuming that https://github.com/nextcloud/helm/pull/379 is merged, this would be the next version up.

PR #321, opened 2022-12-10
Update PR template to not include bumping the chart version

provokateurin:
# Pull Request

## Description of the change

Removes the requirement for bumping the Chart version in PRs.

## Benefits

Collecting multiple PRs before making a new release and also avoid rebasing PRs so often.

## Possible drawbacks

None

## Additional information

Discussed in the [NC Containers public chat](https://cloud.nextcloud.com/call/oefq5xvp)  with @tvories 

## Checklist 

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,5 @@
+<!-- Please do not bump the Chart version! -->
+
 # Pull Request
 
 ## Description of the change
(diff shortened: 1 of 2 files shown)


tvories:
This will remove the requirement to set the chart version, but will that prevent the workflow from trying to release when a PR is merged?

provokateurin:
Would that even work? AFAICT chart-releaser creates the tag and the github release already

tvories:
I am sure there is a modification to our workflow that we can do to allow us to only publish a chart release on a tag.  I think that's all built into the way we currently have it set up, though, so I am thinking we will need to revamp the whole process before we can remove the chart bump requirement

PR #326, opened 2022-12-23
Bug: Fix issue cause by boolean without quotes into env variable

varet80:
The new introduction of env variable, breaks the deployment

this is a helmchart bug fix


provokateurin:
Good catch!

provokateurin (approved):

PR #338, opened 2023-01-24
Add documentation for `primary.persistence.existingClaim` for mariadb and postgresql

### LINKED ISSUE #281: ExistingClaim for MariaDB and Postgresql
If my understanding of the DB volumes is correct, the data that is stored is not just a cache or temporary, thus having more control over the volumes would allow for better backup/migration strategies. It would be useful if the database persistence settings in the chart allowed for the specifying an existingClaim, just like for nextcloud(-data) volumes.

jessebot:
# Pull Request

## Description of the change
This is just documenting already present and working features:

- Adds a commented option for both mariadb and postgresql for existingClaim
- Updates README.md with same options

## Benefits

Prevent people from having to cross-reference multiple charts' READMEs.

## Possible drawbacks

Think we're good to update the readme and comments without any drawbacks :)

## Applicable issues
- fixes #281 

## Additional information

Don't think we need to update the Chart Version for this, but updating the chart version next time there's a change to the actual code, will update the `values.yaml` that outputs when you do `helm show values nextcloud/nextcloud`. In the meantime, it's just a couple of comments, so a minor chart version doesn't even seem necessary.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.4.1
+version: 3.4.2
 appVersion: 25.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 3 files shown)


provokateurin (approved):

provokateurin:
You always need to bump the chart version, otherwise LGTM.

Issue #250, opened 2022-06-28
Release notes for chart


hyperbolic2346:
Am I missing where to find the release notes for the chart? I'm looking to upgrade from 2.14.6 to 3.0.0, but I don't see anything about what has changed and what might break. Are there any configuration differences in these releases?

jessebot:
I don't believe we have release notes right now, but this is something we should be able to automate going into the releases according to this guide:

https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

I can start a PR with a basic starting point for `.github/release.yml`:

```yaml
# .github/release.yml

changelog:
  exclude:
    labels:
      - ignore-for-release
  categories:
    - title: Breaking Changes 🛠
      labels:
        - Semver-Major
        - breaking-change
    - title: Exciting New Features 🎉
      labels:
        - Semver-Minor
        - enhancement
    - title: Other Changes
      labels:
        - "*"
```

I need to test it, and it wouldn't help with previous releases, but it would at least make it so that going forward, you could look at the associated PRs.

In the meantime, each change should have an associated PR, so you can go to the Pull Requests and sort by recently closed, to get an idea of what's changed . From `2.14.6` to `3.0.0` there's been a lot of changes that could be considered breaking. One of the most notable changes will definitely be the appVersion (which is the nextcloud container version tag), so chart version `2.14.6` used the `23.0.3` nextcloud image tag whereas `3.0.0` uses the `24.0.2` image tag. The current nextcloud image tag in `3.4.3` is `25.0.3`, so that's two major versions.

According to the docker repo README right now, it's recommended to only upgrade one major version at a time, so if your docker containers is version 23, you'd need to first upgrade to 24 (so upgrade to chart version 3.0.0), and then to 25:

https://github.com/nextcloud/docker#update-to-a-newer-version

I'm unsure of specific changes to the helm chart itself that would break anything though, I'm going to recommend taking backups of both your files and database regardless, just in case.

@tvories or @provokateurin is there anything either of you can recall that stands out as having been changed that would need revision from the standpoint of the end user values.yaml from chart version 2.14.6 to 3.0.0 outside of the nextcloud container version? To my knowledge, much of it has been just adding features, but I can't recall if anything moved around to different namespaces of the yaml. I may also just be missing in the github interface how to compare a repo tag to the master branch.

jessebot:
[...]

PR #339, opened 2023-01-26
[Draft] Release notes automation

jessebot:
# Pull Request
Currently in draft while I test out the changes listed in https://github.com/nextcloud/helm/pull/321#issuecomment-1405154714

## Description of the change

This change would begin automation of the release notes that go to the descriptions here:
https://github.com/nextcloud/helm/releases

It utilizes the method listed in github's docs here:
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

## Benefits

This will start adding the PR(s) associated with a given release here. Here's an example from one of my repos of how it would look: https://github.com/jessebot/onboardme/releases/tag/v0.16.6

## Possible drawbacks

It's not listed whether or not this applies to all releases, or just releases manually generated through the github UI, so further testing may be needed, and it may be better to do this as a helm chart-releaser step as listed in the comment below, https://github.com/nextcloud/helm/pull/339#issuecomment-1405154600

## Applicable issues

- helps issues similar to #250 in the future.

## Additional information

The [helm/chart-releaser-action](https://github.com/helm/chart-releaser-action) uses [helm/chart-releaser](https://github.com/helm/chart-releaser) which does mention [creating github releases](https://github.com/helm/chart-releaser#create-github-releases-from-helm-chart-packages), however, it comes with some caveats discussed in https://github.com/nextcloud/helm/pull/339#issuecomment-1405154600 and https://github.com/nextcloud/helm/pull/321#issuecomment-1405154714 

## Checklist 

- [] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,7 @@
+# .github/release.yml
+
+changelog:
+  categories:
+    - title: Changes
+      labels:
+        - "*"
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.4.3
+version: 3.4.4
 appVersion: 25.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:


provokateurin:
Please note that every release will only contain a single PR, because the version bump is required for each PR thus creating a new release after every PR is merged. Also see https://github.com/nextcloud/helm/pull/321, but that one is broken :/

PR #341, opened 2023-01-26
update release branch from master to main

### LINKED ISSUE #340: Update default branch name from `master` to `main`
I think if we're going to update our release process a little bit anyway, it might be a good time to finally rename the default branch from `master` to `main` to match modern industry standards. For instance, GitHub has already moved from using master, and has a little guide here:
https://github.com/github/renaming
and here:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch

Others who have changed their standards from master to main include with Python, Bitnami, and ISC. Here's a pretty good RFC draft that explains further why it's important:
https://datatracker.ietf.org/doc/html/draft-knodel-terminology-10

We can easily update our release process to point at main instead of master, and github will do a automatic redirect for those who still have it bookmarked. I'm happy to draw up the PR(s) to fix all the instances of master we have in the helm repo. For those who have master cloned locally, we can also have a small section in our README that notes the branch name change and points at the official instructions on how to update locally here:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch#updating-a-local-clone-after-a-branch-name-changes

I don't have the permissions to do this as a collaborator, so I would need an official maintainer to update the name, but after that is done, there is only one location that would need to be updated in this repo and that's this line from `master` to `main`:
https://github.com/nextcloud/helm/blob/5cbbdf193e2fe2a9f17bfdeb26cf5ef2fd1f676b/.github/workflows/release.yaml#L7

jessebot:
# Pull Request

## Description of the change

Updates default release branch from `master` to `main`

## Benefits

Makes the repo more accessible and keeps us in line with industry standards as listed out in #340.

## Possible drawbacks

May be good to wait until we update the official default branch from master to main, so that there is not any confusion. may also be good to hold off on this specific change if we end up using a tag based release method, as it would just introduce more commits that may be confusing to track.

## Applicable issues

- fixes #340 

## Additional information
I've prepped this PR to make it easier to move forward after getting a maintainer to update the branch name, but it will remain in draft until #339 and #321 are resolved.

## Checklist 

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -4,7 +4,7 @@ name: Release Charts
 on:
   push:
     branches:
-      - master
+      - main
     paths-ignore:
       - 'README.md'
       - 'charts/**/README.md'


provokateurin:
I don't have access, but @juliushaertl for example could do it.

juliushaertl:
Shall I just change it or do you still need to prepare something before?

provokateurin:
I think you can change it and then we merge this PR to fix the CI.

juliushaertl:
Done:

Screenshot 2023-04-24 at 12 45 39

provokateurin (approved):

Issue #333, opened 2023-01-19
Liveness checks kills pod while upgrading


remram44:
It is really unclear how you mean for NextCloud to ever upgrade. There are no instructions #221 and automatic upgrade probably can't work since all replicas would race against each other, and pods get killed while doing it because of failing liveness checks.

How can this work?

provokateurin:
Replicas can't race against each other because there is a locking mechanism implemented in the nextcloud docker image (See https://github.com/nextcloud/docker/blob/master/25/apache/entrypoint.sh#L127-L240).
I never had problems with the probes on my infra, but the default values should definitely work on _most_ hardware, so we should change them accordingly. @remram44 If you could experiment with the values until they work we could compare them with the ones from @jessebot and take the most forgiving ones of each of you.

remram44:
I can change `replicas` to 1 and remove the probes before I upgrade then reset it, but it would probably be better if it worked out-of-the-box. I don't know the consequences of an upgrade getting killed half-way, it probably shouldn't be something that happens in normal operations.

It would probably be better to either:

* Have the probes pass while the upgrade is happening. This is a little tricky because the upgrade happens in the Docker image's entrypoint, and no server is running at that point to reply to GET requests
* Use a Helm hook to do the upgrade outside of the deployment proper. This way you don't need probes, and you can fail the helm upgrade if the upgrade fail

About the number of replicas, there is a locking mechanism in the image (primitive -- not sure why it doesn't use `flock`?), but it needs to be enabled by setting `NEXTCLOUD_INIT_LOCK`. This should probably be set in the Helm chart.

remram44:
@provokateurin `NEXTCLOUD_INIT_LOCK` is NOT set by default as per [the README](https://github.com/nextcloud/docker/blob/f69a48834e16e462410a08bcc878e306bef08002/README.md#L142):

> `NEXTCLOUD_INIT_LOCK` (not set by default) Set it to true to enable initialization locking. Other containers will wait for the current process to finish updating the html volume to continue.

jessebot:
> About the number of replicas, there is a locking mechanism in the image (primitive -- not sure why it doesn't use flock?), but it needs to be enabled by setting NEXTCLOUD_INIT_LOCK. This should probably be set in the Helm chart.

@remram44 would you like to create a PR for this change?

Env vars are created here:
[...]

PR #343, opened 2023-01-27
Add nextcloud.initLock to set NEXTCLOUD_INIT_LOCK

remram44:
# Pull Request

## Description of the change

This allows setting `NEXTCLOUD_INIT_LOCK`, see https://github.com/nextcloud/docker/pull/1760

## Benefits

If running multiple replicas, this will avoid them walking over each other when upgrading.

## Possible drawbacks

If an upgrade is aborted in progress, the lock file might be left behind. You will have to manually delete it before another upgrade can be attempted. This is an issue in the Docker image that I filed here: https://github.com/nextcloud/docker/issues/1903

## Applicable issues

- see #333 (not a complete fix)

## Checklist 


charts/nextcloud/templates/_helpers.tpl
```diff
@@ -156,6 +156,10 @@ Create environment variables used to configure the nextcloud container as well a
 - name: NEXTCLOUD_UPDATE
   value: {{ .Values.nextcloud.update | quote }}
 {{- end }}
+{{- if .Values.nextcloud.initLock }}
```
provokateurin, line 159: Shouldn't we just enable it if replicas > 0? I don't see the point in having an option for it, since it will do nothing in a single replica environment and is required for multiple replica environments.

provokateurin (changes requested):

charts/nextcloud/templates/_helpers.tpl
remram44: The number of replicas can be changed at any time though, using `kubectl scale`, or by horizontal pod autoscaler. If the chart makes a decision when you install/upgrade, that decision can become wrong.

provokateurin: I see, then let's add a big warning that the lock should be enabled when having multiple replicas. We could also just always enable it, since it doesn't hurt setups with only a single replica.

remram44: I agree with you there, I just used the same default as the Docker image. There is the possibility of a lockup if the upgrade fails (because the lock file is left behind), I think that's why it's disabled by default, but I have an open PR against the Docker image to avoid that: https://github.com/nextcloud/docker/issues/1903

provokateurin:
So the docker image now uses flock, so it's safe to enable the init lock by default, right?

remram44:
`NEXTCLOUD_INIT_LOCK` was removed by https://github.com/nextcloud/docker/commit/8aabc4b3ead388d297d2f46646e552ad5143ad0e so this is no longer relevant.

PR #349, opened 2023-02-14
Fix namespace selector

Richardds:
# Pull Request

## Description of the change

The current configuration does not work when the monitored service runs outside of the `serviceMonitor`'s namespace.

Other helm charts, such as [bitnami/redis](https://github.com/bitnami/charts/tree/main/bitnami/redis) use `namespaceSelector` to fix this [[1](https://github.com/bitnami/charts/blob/71af6a6b0c6070a4bd9d2e4466c9daf3c8c612ab/bitnami/redis/templates/servicemonitor.yaml#L38)].

## Benefits

The `serviceMonitor` is now able to find the monitored service.

## Additional information

Example of `metrics` section from `values.yaml`:
```yaml
...
metrics:
  enabled: true
  token: <token>
  serviceMonitor:
    enabled: true
    namespace: monitoring
    labels:
      release: prometheus
```

## Checklist 


Richardds:
Hello @jessebot , I bumped the version to `3.5.1`.

jessebot (approved):
Looks good to me

charts/nextcloud/templates/metrics-servicemonitor.yaml
```diff
@@ -19,6 +19,9 @@ metadata:
     {{- end }}
 spec:
   jobLabel: {{ .Values.metrics.serviceMonitor.jobLabel | quote }}
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace }}
```
jessebot, line 24: Haven't merged this because it doesn't seem to match the other calls to namespaces above:

https://github.com/nextcloud/helm/blob/b7b34cb7ced883ac8e21944255e76cfbffbcec1d/charts/nextcloud/templates/metrics-servicemonitor.yaml#L6-L10

Should we update them to both match?

jessebot (changes requested):

charts/nextcloud/templates/metrics-servicemonitor.yaml
jessebot: As in, have your change also have the if/else, but also make sure to have the ` | quote` bit.

Richardds: Hi @jessebot , I adjusted the code as suggested.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.3
+version: 3.5.4
```
```suggestion
version: 3.5.13
```

charts/nextcloud/templates/metrics-servicemonitor.yaml
jessebot: waiting for the linter to run on this.

charts/nextcloud/templates/metrics-servicemonitor.yaml
```diff
@@ -19,6 +19,13 @@ metadata:
     {{- end }}
 spec:
   jobLabel: {{ .Values.metrics.serviceMonitor.jobLabel | quote }}
+  namespaceSelector:
+    matchNames:
+      {{- if .Values.metrics.serviceMonitor.namespace }}
+      - {{ .Values.metrics.serviceMonitor.namespace | quote }}
+      {{- else }}
+      - {{ .Release.Namespace | quote }}
+      {{- end }}
```
jessebot, line 28: with this whole block, it might need to be moved left two spaces to line up with `matchNames`.
```yaml
    matchNames:
    {{- if .Values.metrics.serviceMonitor.namespace }}
    - {{ .Values.metrics.serviceMonitor.namespace | quote }}
    {{- else }}
    - {{ .Release.Namespace | quote }}
    {{- end }}
```

This would match the way we're doing the gotemplates everywhere else like:
```yaml
    - port: metrics
      path: "/"
      {{- if .Values.metrics.serviceMonitor.interval }}
      interval: {{ .Values.metrics.serviceMonitor.interval }}
      {{- end }}
      {{- if .Values.metrics.serviceMonitor.scrapeTimeout }}
      scrapeTimeout: {{ .Values.metrics.serviceMonitor.scrapeTimeout }}
      {{- end }}
```

It's been a long week, @provokateurin can you just confirm I'm making sense here? I don't have time to test the helm chart directly for this specific issue right this second, but I will be tackling metrics probably in another month or two for a work project.

charts/nextcloud/templates/metrics-servicemonitor.yaml
```diff
@@ -19,6 +19,13 @@ metadata:
     {{- end }}
 spec:
   jobLabel: {{ .Values.metrics.serviceMonitor.jobLabel | quote }}
+  namespaceSelector:
+    matchNames:
+    {{- if .Values.metrics.serviceMonitor.namespace }}
```
wrenix, line 24:
```suggestion:+2
    {{- with .Values.metrics.serviceMonitor.namespace }}
    - {{ . | quote }}
```

wrenix (changes requested):
I miss a documentation in the `values.yaml`

charts/nextcloud/templates/metrics-servicemonitor.yaml
wrenix: 1. with is easy reable
2. so multiple namespaces are configurable

charts/nextcloud/templates/metrics-servicemonitor.yaml
jessebot: I missed this ages ago and now your other PR makes sense :) this was also before I realized you have a bit more freedom when using gotemplating :)

charts/nextcloud/templates/metrics-servicemonitor.yaml
jessebot: agree

wrenix:
i believe we could close this PR now ;)
#524 was merged

provokateurin:
You're right

PR #351, opened 2023-02-18
Separate resources for cronjob sidecar

jilleJr:
# Pull Request

## Description of the change

Changed so cronjob container uses `.Values.cronjob.resources` instead of `.Values.resources`

## Benefits

The cronjob sidecar takes way less resources than the main NextCloud container, but they still used the same resource requests/limits.

With this, you can properly give the cronjob container tiny amount of RAM and CPU, and host more pods on the same node.

## Possible drawbacks

If someone has some automatic checks in their cluster, like with OPA, to reject containers without resource requests/limits, then this would fail for them. Small nuesense.

Could do `.Values.cronjob.resources | default .Values.resources` to keep the current behaviour, but I think having them disconnected is a better default.

## Applicable issues

- fixes #

## Checklist 


jessebot (approved):
This looks good to me. Thank you for your contribution :)

PR #277, opened 2022-09-06
Fix: extraVolumeMounts for Nginx Container

Nold360:
# Pull Request

## Description of the change

Add `extraVolumeMounts` to nginx container. `extraVolumes` are already applied to the container, just the mount is missing.

## Benefits

Lets you add extra volumes to nginx container.

## Possible drawbacks

Mounting volumes you might not want on the webserver?!

## Applicable issues

- fixes #

## Checklist 

- [x] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] (optional) Variables are documented in the README.md

PR #356, opened 2023-03-02
Add NGINX metrics as an option

Jeroen0494:
# Pull Request

## Description of the change

* #277 introduced an erroneous YAML file, where the included YAML is not on a new line but pasted after the previous. This PR fixes this by using the same code as used for the Nextcloud container.
```
        - name: nextcloud-nginx-config
          mountPath: /etc/nginx/nginx.conf
          subPath: nginx.conf
          readOnly: true        - mountPath: /var/www/html/data/XXX   <-- this is wrong
          name: nextcloud-data-XXX
```

* Add additional securityContext options for the metrics exporter.
* The metrics exporter hostname was not included in the trusted domains, which makes the exporter fail to collect data.
* Add the option for including NGINX metrics collection. The allow rules are for a default k3s installation and service just as an example.

## Benefits

* volumeMounts are now fixed.
* securityContext makes my polaris dashboard (and kube-bench) happy.
* Metrics can not be collected properly.
* Metrics for NGINX are collected and automatically exported by the service monitor. You can use a Grafana dashboard to monitor connections.

## Possible drawbacks

* None, it's optional

## Applicable issues

- fixes #277

## Checklist 


charts/nextcloud/README.md
```diff
@@ -202,6 +202,9 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `metrics.image.repository`                                   | Nextcloud metrics exporter image name                                                  | `xperimental/nextcloud-exporter`                             |
 | `metrics.image.tag`                                          | Nextcloud metrics exporter image tag                                                   | `0.5.1`                                                      |
 | `metrics.image.pullPolicy`                                   | Nextcloud metrics exporter image pull policy                                           | `IfNotPresent`                                               |
+| `metrics.nginx.enabled`                                      | Start NGINX metrics configuration                                                      | `false`                                                      |
+| `metrics.nginx.allow`                                      | NGINX metrics configuration allow list                                                 | not set                                                      |
```
jessebot, line 206: Could you please fix the markdown table here?

charts/nextcloud/README.md
```diff
@@ -202,6 +202,9 @@ The following table lists the configurable parameters of the nextcloud chart and
...
+| `metrics.nginx.service.port`                                 | NGINX metrics configuration service port                                               | `8081`                                                      |
```
jessebot, line 207: also here for fixing markdown table

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.2
+version: 3.5.3
```
jessebot, line 3: This is also out of date and would need to be bumped further

jessebot (changes requested):

Jeroen0494:
Rebased on current master and fixed markdown table.

Jeroen0494:
Uhm, I think my request for re-review removed @tvories and @provokateurin from the requested reviewers. Sorry.

charts/nextcloud/README.md
```diff
@@ -52,7 +52,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | ------------------------------------------------------------ | -------------------------------------------------------------------------------------- | ------------------------------------------------------------ |
 | `image.repository`                                           | nextcloud Image name                                                                   | `nextcloud`                                                  |
 | `image.flavor`                                               | nextcloud Image type (Options: apache, fpm)                                            | `apache`                                                     |
-| `image.tag`                                                  | nextcloud Image tag                                                                    | `{VERSION}`                                                  |
+| `image.tag`                                                  | nextcloud Image tag                                                                    | `{VERSION}`                                                            |
```
provokateurin, line 55: Not sure what's up with those random re-indentations, please remove.

charts/nextcloud/templates/_helpers.tpl
```diff
@@ -151,7 +151,11 @@ Create environment variables used to configure the nextcloud container as well a
       name: {{ .Values.nextcloud.existingSecret.secretName | default (include "nextcloud.fullname" .) }}
       key: {{ .Values.nextcloud.existingSecret.passwordKey | default "nextcloud-password" }}
 - name: NEXTCLOUD_TRUSTED_DOMAINS
+{{- if .Values.metrics.enabled }}
+  value: {{ .Values.nextcloud.host }} metrics-server
```
provokateurin, line 155: Does this space separation work?

charts/nextcloud/templates/deployment.yaml
```diff
@@ -238,12 +255,17 @@ spec:
...
+        {{- if .Values.metrics.nginx.enabled }}
+        - name: nginx-metrics-config
+          mountPath: /etc/nginx/conf.d/metrics.conf
+          readOnly: true
+          subPath: metrics.conf
+        {{- end }}
         - name: nextcloud-nginx-config
           mountPath: /etc/nginx/nginx.conf
           subPath: nginx.conf
-        {{- if .Values.nextcloud.extraVolumeMounts }}
```
provokateurin, line 244: Why is this removed and nextcloud.volumeMounts added? It adds a bunch of volume mounts that are not needed for the nginx container, so why would you include them?

charts/nextcloud/templates/deployment.yaml
```diff
@@ -325,6 +347,20 @@ spec:
       - name: nextcloud-nginx-config
         configMap:
           name: {{ template "nextcloud.fullname" . }}-nginxconfig
+          defaultMode: 420
+          items:
+            - key: nginx.conf
+              mode: 292
+              path: nginx.conf
```
provokateurin, line 354: Was this broken before?

charts/nextcloud/templates/deployment.yaml
```diff
@@ -201,6 +204,20 @@ spec:
...
+            port: http
+            httpHeaders:
+            - name: Host
+              value: {{ .Values.nextcloud.host | quote }}
+          initialDelaySeconds: {{ .Values.startupProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.startupProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.startupProbe.timeoutSeconds }}
+          successThreshold: {{ .Values.startupProbe.successThreshold }}
+          failureThreshold: {{ .Values.startupProbe.failureThreshold }}
+        {{- end }}
```
provokateurin, line 220: Is this required for the metrics or did you just add it because it was missing?

charts/nextcloud/values.yaml
```diff
@@ -465,6 +477,14 @@ metrics:
 
   # podLabels: {}
 
+  nginx:
+    enabled: false
+    allow:
+    - 10.233.105.0/24
+    - 10.43.0.0/16
```
provokateurin, line 484: What are those?

provokateurin (changes requested):

charts/nextcloud/templates/deployment.yaml
Jeroen0494: Because I added the metrics configuration to the existing configuration file, I need to specify both via items instead of just mounting the entire ConfigMap. Otherwise metrics will be mounted as well, even though it may not have been enabled.
So not broken, but now necessary to specify which keys to mount.

charts/nextcloud/templates/deployment.yaml
Jeroen0494: I added it because is was missing. The Nextcloud container did have it specified.

charts/nextcloud/values.yaml
Jeroen0494: Those serve just as an example (default k3s installation networks). I can comment them out if preferred.

charts/nextcloud/templates/_helpers.tpl
Jeroen0494: Yes it does:
```
        - name: NEXTCLOUD_TRUSTED_DOMAINS
          value: nextcloud.mediaserver.fritz.box metrics-server
```

Because Kubernetes assumes it's a string. But I can add double quotes for safety.

charts/nextcloud/values.yaml
provokateurin: Yeah I think that's better

charts/nextcloud/README.md
Jeroen0494: My screen isn't wide enough to fully display the table and catch all of these, will fix in a while.

charts/nextcloud/templates/_helpers.tpl
provokateurin: I'd prefer that

charts/nextcloud/templates/deployment.yaml
Jeroen0494: I had the fix for https://github.com/nextcloud/helm/pull/358 in this PR first, but then that one got merged, and then I rebased. Now I think something may have gone wrong during rebase.
It was my intention not to have an if statement, because that's how it's specified for the Nextcloud container as well. But I should have used `extraVolumeMounts` instead of `volumeMounts`.

Would you prefer I revert this, of change my version to `extraVolumeMounts`?

provokateurin: Better to revert it and try to not get unrelated changes in the PR.

Jeroen0494: Done

Jeroen0494:
Rebased on current master branch.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.8
+version: 3.5.9
```
```suggestion
version: 3.5.10
```

charts/nextcloud/templates/_helpers.tpl
```diff
@@ -151,7 +151,11 @@ Create environment variables used to configure the nextcloud container as well a
       name: {{ .Values.nextcloud.existingSecret.secretName | default (include "nextcloud.fullname" .) }}
       key: {{ .Values.nextcloud.existingSecret.passwordKey | default "nextcloud-password" }}
 - name: NEXTCLOUD_TRUSTED_DOMAINS
+{{- if .Values.metrics.enabled }}
+  value: "{{ .Values.nextcloud.host }} metrics-server"
+{{- else }}
```
jessebot, line 156: This looks like it doesn't include the main nextcloud host name, if you have the metrics-server enabled. Shouldn't we have both?

charts/nextcloud/values.yaml
```diff
@@ -194,10 +194,15 @@ nextcloud:
...
-  #   runAsGroup: 33
-  #   runAsNonRoot: true
-  #   readOnlyRootFilesystem: false
+    # allowPrivilegeEscalation: false
+    # capabilities:
+    #   drop:
+    #     - ALL
+    # privileged: false
+    # readOnlyRootFilesystem: true
+    # runAsGroup: "33"
```
jessebot, line 203:
```suggestion
    # www-data user for all nextcloud containers is UID 33, except tags with alpine in them which use UID 82
    # runAsGroup: 33
```
set value to int64 vs string, add comment for clarity

charts/nextcloud/values.yaml
```diff
@@ -194,10 +194,15 @@ nextcloud:
...
+    # runAsNonRoot: true
+    # runAsUser: "33"
```
jessebot, line 205:
```suggestion
    # runAsUser: 33
```
set value to int64 from string

charts/nextcloud/values.yaml
jessebot: wait, didn't this get fixed in another PR already? 🤔

jessebot: ah, yeah, it was in this https://github.com/nextcloud/helm/pull/378/files - need to rebase then

charts/nextcloud/templates/_helpers.tpl
Jeroen0494: It does, see here:
https://github.com/nextcloud/helm/pull/356#discussion_r1133241859

It's not an OR statement.

Jeroen0494:
@jessebot thanks for the review. The quotes for user and group numbers were my mistake. I also updated the securityContext, but didn't see the removed quoted from upstream.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -173,6 +173,9 @@ spec:
         - name: http
           containerPort: {{ .Values.nextcloud.containerPort | default "80" }}
           protocol: TCP
+        - name: metrics
+          containerPort: {{ .Values.metrics.nginx.service.port | default "8081" }}
```
wrenix, line 177: please do not put defaults into templates - therefore `values.yaml` exists

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -9,6 +9,36 @@ metadata:
...
+        add_header X-XSS-Protection                     "1; mode=block" always;
+
+        # NGINX metrics
+        location /stub_status {
+            stub_status on;
+            allow 127.0.0.1;
+            {{- range .Values.metrics.nginx.allow }}
+                allow {{ . }};
+            {{- end }}
+            deny all;
```
wrenix, line 38: it would be nice, that `deny all` could also be disabled (maybe if no allow exists ...)

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -44,7 +74,7 @@ data:
         }
 
         server {
-            listen 80;
+            listen {{ .Values.nextcloud.containerPort | default "80" }};
```
wrenix, line 77: default should be in `values.yaml`

charts/nextcloud/templates/service.yaml
```diff
@@ -21,6 +21,12 @@ spec:
     {{- if eq .Values.service.type "NodePort" }}
     nodePort: {{ default "" .Values.service.nodePort}}
     {{- end }}
+  {{- if and .Values.nginx.enabled .Values.metrics.nginx.enabled }}
+  - name: metrics
+    port: {{ .Values.metrics.nginx.service.port | default "8081" }}
```
wrenix, line 26: default should be in `values.yaml`

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.9
+version: 3.5.10
```
wrenix, line 3: semver -> this is a feature, so it should be 3.6.0

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -9,6 +9,36 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
+{{- if .Values.metrics.nginx.enabled }}
+  metrics.conf: |
+    server {
+        listen {{ .Values.metrics.nginx.service.port | default "8081" }};
```
wrenix, line 15: default should be in `values.yaml`

charts/nextcloud/Chart.yaml
jessebot: It would be a bit higher than this now, as the chart has moved forward a bit since this version.

Tangentially: I'd like to be able to label issues as major, minor, or patch versions and then a GitHub Action can just keep the PRs up to date one major, minor, or patch version as compared to the `main` branch. I will probably create an issue for that, because it would solve a lot of merge conflicts in a lot of PRs that may be slow moving because of that.

Jeroen0494:
I've rebased as best as possible.

wrenix:
It looks nice, thank you for your work.

Do you like to write simething for a sidecar with an nginx-exporter and maybe adjust or add an ServiceMonitor for Prometheus?

Jeroen0494:
> It looks nice, thank you for your work.
> 
> Do you like to write simething for a sidecar with an nginx-exporter and maybe adjust or add an ServiceMonitor for Prometheus?
> 
> nginx exporter:
> 
> * from nginx itself: https://github.com/nginxinc/nginx-prometheus-exporter
> * use telegraf as exporter (input: [nginx](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/nginx), output: [prometheus_client](https://github.com/influxdata/telegraf/tree/master/plugins/outputs/prometheus_client))

Prometheus already has their own Helm chart for this, I'd rather not reinvent the wheel. Maybe you could set this as a dependency when metrics is enabled, but for now I'd just like to have this included. This PR has been open for so long already.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -360,6 +360,7 @@ spec:
         - name: nextcloud-nginx-config
           configMap:
             name: {{ template "nextcloud.fullname" . }}-nginxconfig
+            defaultMode: 420
```
jessebot, line 363: This should be configurable in the values.yaml, and the default in values.yaml can be `420` if you'd like.

charts/nextcloud/templates/metrics/deployment.yaml
```diff
@@ -73,6 +73,12 @@ spec:
...
           {{- end }}
           securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
             runAsUser: 1000
             runAsNonRoot: true
```
jessebot, line 83:
```suggestion
          {{- with .Values.metrics.securityContext }}
          securityContext:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

We should have this whole block be configurable based on the metrics securityContext in values.yaml.

charts/nextcloud/values.yaml
```diff
@@ -513,6 +525,15 @@ metrics:
 
   # podLabels: {}
```
jessebot, line 527:
```suggestion
  # Set metrics container securityContext parameters.
  securityContext:
    runAsUser: 1000
    runAsNonRoot: true
```

We should make the metrics pod security context configurable something like in this suggestion in the values.yaml, as mentioned above.

jessebot (changes requested):
Added a few more notes as there's some defaults that have been set in the templates that should be configurable.

charts/nextcloud/templates/nginx-config.yaml
jessebot: I agree with @wrenix on this point.

Jeroen0494:
I've included your comments and squashed my commits to make everything more legible. Ready for another review.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -350,16 +350,25 @@ spec:
         - name: nextcloud-config
           configMap:
             name: {{ template "nextcloud.fullname" . }}-config
+            {{- if .Values.nextcloud.configs.defaultMode }}
```
wrenix, line 353: Please use `with`

```suggestion
            {{- with .Values.nextcloud.configs.defaultMode }}

            defaultMode: {{ . }}
            {{- end }}
```

PS: your current code would not work

charts/nextcloud/templates/deployment.yaml
```diff
@@ -350,16 +350,25 @@ spec:
...
+            defaultMode: .Values.nextcloud.configs.defaultMode
+            {{- end }}
         {{- end }}
         {{- if .Values.nextcloud.phpConfigs }}
         - name: nextcloud-phpconfig
           configMap:
             name: {{ template "nextcloud.fullname" . }}-phpconfig
+            {{- if .Values.nextcloud.configs.defaultMode }}
+            defaultMode: .Values.nextcloud.configs.defaultMode
```
wrenix, line 362: that does not work, it is missing `{{`.

Please use `with` in statt of `if`

charts/nextcloud/values.yaml
```diff
@@ -128,6 +128,7 @@ nextcloud:
   # Extra config files created in /var/www/html/config/
   # ref: https://docs.nextcloud.com/server/15/admin_manual/configuration_server/config_sample_php_parameters.html#multiple-config-php-file
   configs: {}
+    # defaultMode: 420
```
wrenix, line 131: Please put a line without outcommanting (`if` and `with` would handle it as an empty/false)
```suggestion
    # set defaultMode for mounted configMaps (e.g. defaultMode: 420)
    defaultMode:
```

charts/nextcloud/values.yaml
```diff
@@ -513,6 +526,26 @@ metrics:
 
   # podLabels: {}
 
+  nginx:
+    enabled: false
+    allow: []
+    # Example
+    # - 10.233.105.0/24
+    # - 10.43.0.0/16
+    service:
```
wrenix, line 535: This value has no affact anywhere ....
But maybe usefull (on deployment under ports and your ConfigMap rendered; or service).

wrenix (changes requested):
Still something todo

charts/nextcloud/values.yaml
Jeroen0494: This is actually used in nginx-config.yaml:

```
            {{- range .Values.metrics.nginx.allow }}
                allow {{ . }};
            {{- end }}
```

Jeroen0494:
Fixes included, I think.

charts/nextcloud/templates/deployment.yaml
jessebot: github is beign weird and won't let me merge this comment, so I'm commenting lower

charts/nextcloud/templates/deployment.yaml
```diff
@@ -350,16 +350,25 @@ spec:
         - name: nextcloud-config
           configMap:
             name: {{ template "nextcloud.fullname" . }}-config
+            {{- if .Values.nextcloud.configs.defaultMode }}
+            defaultMode: .Values.nextcloud.configs.defaultMode
+            {{- end }}
```
jessebot: As @wrenix said, this could be `with` and it would save on required code, also you need `{{}}` surrounding the `.Values.nextcloud.configs.defaultMode` or the helm template won't work render:

```suggestion
            {{- with .Values.nextcloud.configs.defaultMode }}
            defaultMode: {{ . }}
            {{- end }}
```

You can commit this suggestion directly.

charts/nextcloud/templates/deployment.yaml
jessebot: echoing wrenix here since it won't let me make a suggestion in a comment reply to their comment.

```suggestion
            {{- with .Values.nextcloud.configs.defaultMode }}
            defaultMode: {{ . }}
```

You can commit this suggestion directly from this PR.

jessebot: resolving in favor of my comment with the suggestion, but echoing your same sentiment :pray:

charts/nextcloud/templates/deployment.yaml
```diff
@@ -350,16 +350,25 @@ spec:
...
         {{- end }}
         {{- if .Values.nginx.enabled }}
         - name: nextcloud-nginx-config
           configMap:
             name: {{ template "nextcloud.fullname" . }}-nginxconfig
+          {{- if .Values.nextcloud.configs.defaultMode }}
+            defaultMode: .Values.nextcloud.configs.defaultMode
```
jessebot:
```suggestion
          {{- with .Values.nextcloud.configs.defaultMode }}
            defaultMode: {{ . }}
```

charts/nextcloud/values.yaml
jessebot: `.Values.metrics.nginx.allow` is not the same as `.Values.nginx.metrics.service.port`, which is the values you have here. Am I maybe misunderstanding something?

jessebot (changes requested):
Requesting a few changes to make the templates render properly and reduce wordiness.

PR #359, opened 2023-03-04
Update docs to have sections and add existingSecret for mariadb, postgresql, and redis

jessebot:
# Pull Request

## Description of the change

Adding `exsitingSecret` parameters for MariaDB, PostgreSQL, and Redis.

Also updating docs to:
-  make it easier to link to sections of the configuration by separating out:
  - database config params
    - added new exsitingSecret params for both mariadb/postgres
  - persistence config params
  - metrics config params
  - add redis existingSecret params
- update supported versions of k8s (we can't support 1.10-1.19 as they are [not even in support by k8s anymore](https://kubernetes.io/releases/))
- update/add links to bitnami helm charts (they use main instead of master branch now)

## Benefits

Gives README a much needed face lift to make it easier to find stuff. Allows people to use existing secrets by default for all our services.

## Possible drawbacks

🤷 

## Applicable issues

- closes #350

## Checklist 

- [x] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] (optional) Variables are documented in the README.md

PR #366, opened 2023-03-26
Set DB name and host via external secret

jacksgt:
Hi everyone, thanks for maintaining this useful Helm chart!

I'm using the [DB operator](https://github.com/kloeckner-i/db-operator/blob/master/docs/creatingdatabases.md) to automatically provision databases within my cluster.
It works very similar to Cert-Manager: you create a `Database` CR (instead of a `Certificate` CR) and the operator generates a Secret that contains the connection details for the database, like this:

```yaml
apiVersion: v1
kind: Secret
metadata:
  labels:
    created-by: db-operator
  name: nextcloud-db-credentials
type: Opaque
data:
  DB: << base64 encoded database name (generated by db operator) >>
  PASSWORD: << base64 encoded password (generated by db operator) >>
  USER: << base64 encoded user name (generated by db operator) >>
  HOST: << base64 encoded hostname (generated by db operator) >>
```

For this purpose, I would like to be able to also inject external database hostnames and names via a Secret (instead of passing them through the chart parameters).
This patch enables that functionality.

### DIFF
--- a/charts/nextcloud/templates/_helpers.tpl
+++ b/charts/nextcloud/templates/_helpers.tpl
@@ -110,8 +110,21 @@ Create environment variables used to configure the nextcloud container as well a
 {{- else }}
   {{- if eq .Values.externalDatabase.type "postgresql" }}
 - name: POSTGRES_HOST
+  {{- if .Values.externalDatabase.existingSecret.hostKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-%s" .Release.Name "db") }}
+      key: {{ .Values.externalDatabase.existingSecret.hostKey }}
+  {{- else }}
   value: {{ .Values.externalDatabase.host | quote }}
+  {{- end }}
 - name: POSTGRES_DB
+  {{- if .Values.externalDatabase.existingSecret.databaseKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-%s" .Release.Name "db") }}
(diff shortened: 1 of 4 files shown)


jessebot (dismissed):
Could you please rebase your feature branch? We've had a major change the README and some additional values have been added to values.yaml in [this merged PR](https://github.com/nextcloud/helm/pull/359/files). We also cannot merge anything without the DCO check passing. After you've rebased, signed off on all commits, and updated your helm chart version, we can take another look :)

jacksgt:
Hi @jessebot , I rebased the PR as requested.

jacksgt:
@jessebot Done.

charts/nextcloud/templates/_helpers.tpl
```diff
@@ -110,8 +110,21 @@ Create environment variables used to configure the nextcloud container as well a
...
   value: {{ .Values.externalDatabase.host | quote }}
+  {{- end }}
 - name: POSTGRES_DB
+  {{- if .Values.externalDatabase.existingSecret.databaseKey }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-%s" .Release.Name "db") }}
+      key: {{ .Values.externalDatabase.existingSecret.databaseKey }}
+  {{- else }}
   value: {{ .Values.externalDatabase.database | quote }}
```
jessebot, line 128:
```suggestion
  value: {{ .Values.externalDatabase.database | quote }}
  {{- end }}
```
I think this will fix the error here:
https://github.com/nextcloud/helm/actions/runs/5530735352/jobs/10090512180?pr=366

jessebot (approved):
All right, this looks good now! :) Thank you for the feature!

PR #369, opened 2023-04-01
Clarify the behavior of "internalDatabase" setting

jggc:
Clarify that internalDatabase must be enabled: false to activate other databases in nextcloud config. Simply enabling mariadb, external or postgres database will not actually configure nextcloud to use it unless internalDatabase.enabled is false.

# Pull Request

## Description of the change

Only some clarifying comments

## Benefits

Saves some confusion that requires newcomers to go and read the templates to understand what the internalDatabase setting actually does

## Possible drawbacks

Adds 4 lines of comments in values.yaml, makes for a longer file, uses more space in the git repo and every clone. Should not cause too many issues in this world where terabytes of data is cheap.

## Applicable issues

- fixes my nextcloud deployment that had a mariadb database running but was still using the sqlite database since I thought that "internalDatabase" could mean a database controlled by the helm chart, since "externalDatabase" means an external mysql/postgres database running outstide of the chart.

## Additional information

I'm kind of having fun writing way too much information about this 4 line comment only change. But if you want some additional information I think I may take the time to rename the "internalDatabase" setting to something that depicts more obviously that this setting is the master setting that controls the sqlite database. Or better yet, have a setting that makes more explicit that it is controlling nextcloud database connection configuration.

## Checklist 

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -228,7 +228,12 @@ nginx:
   #   runAsNonRoot: true
   #   readOnlyRootFilesystem: true
 
+##
+## Internal sqlite database configuration
+##
 internalDatabase:
+  ## Wether to use the sqlite database. This takes precedence on other database settings
+  ## and must be false to use any other database type
   enabled: true
   name: nextcloud
 


provokateurin:
Looks good, but I don't think you need to bump the chart version, so it's better to leave it as it only adds some comments.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.11
+version: 3.5.12
```
```suggestion
version: 3.5.11
```
change chart version back as per kate's advice, since this is just comments

provokateurin:
@jessebot I don't have permissions to force merge. So the bump is required :/ Sorry for the noise

charts/nextcloud/Chart.yaml
jessebot:
```suggestion
version: 3.5.13
```

jessebot: Can't do a sign off and commit on an old commit.

@jggc if you can bump the chart version , we can get this merged. Sorry for the confusion.

PR #371, opened 2023-04-10
[feat]: support configuration of loadBalancerClass and externalTrafficPolicy

volker-raschek:
# Pull Request

## Description of the change

This PullRequest extends the configuration of the k8s service for nextcloud setups running with metallb on prem. The problem is, if multiple k8s nodes are running the sourceIP of the client will be masked by the ip of the k8s nodes, because the traffic will be forwarded - node hopping. To avoid this behavior the externalTrafficPolicy can be set to Local. This property is currently missing. 

Furthermore, if on prem are multiple loadbalancers are installed it is not possible to specify explicit a different as the default.

## Benefits

The bruteforce protection mechanism works for on prem installations with metallb correctly, if externalTrafficPolicy is set to Local. Otherwise will be blocked the NodeIP and all users are affected by blocking one user.

## Documentation

https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport
https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-loadbalancer

### DIFF
--- a/charts/nextcloud/templates/service.yaml
+++ b/charts/nextcloud/templates/service.yaml
@@ -10,18 +10,18 @@ metadata:
     app.kubernetes.io/component: app
 spec:
   type: {{ .Values.service.type }}
-  {{- if .Values.service.externalTrafficPolicy }}
+  {{- if and .Values.service.externalTrafficPolicy (or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort"))}}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
   {{- end }}
   {{- if eq .Values.service.type "LoadBalancer" }}
   loadBalancerIP: {{ default "" .Values.service.loadBalancerIP }}
-  {{- end }}
-  {{- if and .Values.service.loadBalancerClass (eq .Values.service.type "LoadBalancer") }}
-  loadBalancerClass: {{ .Values.service.loadBalancerClass }}
+  {{- if .Values.service.loadBalancerClass }}
+  loadBalancerClass: {{ default "" .Values.service.loadBalancerClass }}
   {{- end }}
   {{- with .Values.service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges:
   {{- toYaml . | nindent 2 }}
+  {{- end }}
   {{- end }}
   ports:
   - port: {{ .Values.service.port }}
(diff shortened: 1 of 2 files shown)


charts/nextcloud/templates/service.yaml
```diff
@@ -13,6 +13,13 @@ spec:
   {{- if eq .Values.service.type "LoadBalancer" }}
   loadBalancerIP: {{ default "" .Values.service.loadBalancerIP }}
   {{- end }}
+  {{- if and .Values.service.loadBalancerClass (eq .Values.service.type "LoadBalancer") }}
```
provokateurin, line 16: Can you change the order of the statements so that it matches the one above more closely?

charts/nextcloud/templates/service.yaml
```diff
@@ -13,6 +13,13 @@ spec:
...
+  loadBalancerClass: {{ .Values.service.loadBalancerClass }}
+  {{- end }}
+  {{- with .Values.service.loadBalancerSourceRanges }}
```
provokateurin, line 19: Shouldn't there be a check if the value is even set? And also check if the service is of type LoadBalancer.

charts/nextcloud/values.yaml
```diff
@@ -343,6 +343,8 @@ service:
   type: ClusterIP
   port: 8080
   loadBalancerIP: nil
+  # loadBalancerClass:
```
provokateurin, line 346: /
```suggestion
  # loadBalancerClass: nil
```
/

provokateurin (changes requested):

charts/nextcloud/templates/service.yaml
volker-raschek: I modified the conditions and the scope of the if statement

charts/nextcloud/values.yaml
```diff
@@ -341,8 +341,11 @@ cronjob:
 
 service:
   type: ClusterIP
+  # externalTrafficPolicy:
```
provokateurin, line 344:
```suggestion
  # externalTrafficPolicy: nil
```

provokateurin (changes requested):

charts/nextcloud/templates/service.yaml
volker-raschek: Exactly

charts/nextcloud/values.yaml
volker-raschek: Adapted

charts/nextcloud/values.yaml
volker-raschek: Adapted

PR #373, opened 2023-04-10
Update helm/chart-releaser-action `v1.4.1` -> `v1.5`; relax requirements for patch versions

jessebot:
# Pull Request

## Description of the change

This updates the `helm/chart-releaser-action` in our github workflow from `v1.4.1` to `v1.5`. I've also dropped the patch versions, so that we can always make sure we have patch releases up to date, and then we can still be strict when it comes to minor and major releases.

## Benefits

Keeps our workflows up to date, for security reasons and potential deprecations.

## Applicable issues

- fixes #329 

## Additional information

Didn't update chart version, because it shouldn't need to be for the github workflow updates

## Checklist 

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -16,7 +16,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0        
+          fetch-depth: 0
 
       - name: Install Helm
         uses: azure/setup-helm@v3.1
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-test.yaml
```diff
@@ -17,7 +16,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0
+          fetch-depth: 0
```
provokateurin, line 19: Remove spaces

provokateurin (changes requested):

.github/workflows/lint-test.yaml
jessebot: I have no idea where this came from 🤷 but great catch 😄

provokateurin (approved):

Issue #279, opened 2022-09-09
Metrics asking nextcloud trough the ingress instead of local service


millaguie:
The current configuration of metrics sets the NEXTCLOUD_SERVER variable using the ingress URL, this generates extra traffic on the ingress (and logs), and if you are paranoid and don't allow the nodes on the cluster to access the loadbalancer of the ingress prevents you from retrieving the metrics.

I suggest changing this from `metrics-deployment.yaml`
```
        - name: NEXTCLOUD_SERVER
          value: http{{ if .Values.metrics.https }}s{{ end }}://{{ .Values.nextcloud.host }}
```

To something like 
```
        - name: NEXTCLOUD_SERVER
          value: http://{{ template "nextcloud.fullname" . }}:{{ .Values.service.port }}
```

jessebot:
Thanks for bringing this up :) If you can test this and ensure it still works as previously intended, you can submit a PR, and I can review it. I haven't had a chance to test this myself though, so I can't move forward with a PR myself.

tarhann:
We also encounter this issue, please raise the priority on it :)

provokateurin:
@tarhann This chart is community maintained and not an official Nextcloud product, but we of course appreciate PRs with fix and features.

tarhann:
HTH :)

PR #388, opened 2023-04-26
fix(metrics-deployment): use service name and port

tarhann:
# Pull Request

## Description of the change

As stated out by @millaguie, the NEXTCLOUD_SERVER variable of [metrics-deployment](https://github.com/nextcloud/helm/blob/b35eaa38a721542d7ce73d07e123aa0574244c26/charts/nextcloud/templates/metrics-deployment.yaml#L53) should point to the Nextcloud Service name. But that gets its name from the template function `nextcloud.fullname` and not `.Values.nextcloud.host`. 
Also the port was not contained, that both is changed now.

## Benefits

`metrics-exporter`is now able to reach the Nextcloud Service correctly

## Applicable issues

- fixes [#279](https://github.com/nextcloud/helm/issues/279)

### DIFF
--- a/charts/nextcloud/templates/metrics-deployment.yaml
+++ b/charts/nextcloud/templates/metrics-deployment.yaml
@@ -51,7 +51,7 @@ spec:
               key: {{ .Values.nextcloud.existingSecret.passwordKey | default "nextcloud-password" }}
         {{- end }}
         - name: NEXTCLOUD_SERVER
-          value: http{{ if .Values.metrics.https }}s{{ end }}://{{ .Values.nextcloud.host }}
+          value: http{{ if .Values.metrics.https }}s{{ end }}://{{ template "nextcloud.fullname" . }}:{{ .Values.service.port }}
         - name: NEXTCLOUD_TIMEOUT
           value: {{ .Values.metrics.timeout }}
         - name: NEXTCLOUD_TLS_SKIP_VERIFY
--- a/charts/nextcloud/templates/metrics-deployment.yaml
+++ b/charts/nextcloud/templates/metrics-deployment.yaml
@@ -50,6 +50,7 @@ spec:
               name: {{ .Values.nextcloud.existingSecret.secretName | default (include "nextcloud.fullname" .) }}
               key: {{ .Values.nextcloud.existingSecret.passwordKey | default "nextcloud-password" }}
         {{- end }}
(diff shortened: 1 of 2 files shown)


charts/nextcloud/templates/metrics-deployment.yaml
```diff
@@ -51,7 +51,7 @@ spec:
               key: {{ .Values.nextcloud.existingSecret.passwordKey | default "nextcloud-password" }}
         {{- end }}
         - name: NEXTCLOUD_SERVER
-          value: http{{ if .Values.metrics.https }}s{{ end }}://{{ .Values.nextcloud.host }}
+          value: http{{ if .Values.metrics.https }}s{{ end }}://{{ template "nextcloud.fullname" . }}:{{ .Values.service.port }}
```
provokateurin, line 54: Can you add a comment above this line that documents that this points to the service in https://github.com/nextcloud/helm/blob/b35eaa38a721542d7ce73d07e123aa0574244c26/charts/nextcloud/templates/service.yaml#L4 ?

provokateurin (changes requested):

provokateurin:
Please fix DCO and chart version.

charts/nextcloud/templates/metrics-deployment.yaml
provokateurin: I meant more about the resolution of the address of the service (explaining that `template "nextcloud.fullname" .` is used to resolve the service host)

tarhann: Isn't that what k8s-services are generally there for?

provokateurin: Yes, but I wanted a comment like `"nextcloud.fullname" resolves the hostname of the nextcloud service`.

tarhann: I provided a fix for the bug, its up to you what you do with it. I don't see such a comment useful since its obvious where the service gets its name from, what a K8s-Service is for and therefore I wont write it.

jessebot: Adding comments for others that are new to the ecosystem is always a good idea.

FlorentPoinsaut:
Thank you so much @tarhann for this work.
I can confirm that it works very well!

jessebot (approved):
Looks good to me!

provokateurin (approved):

PR #400, opened 2023-05-19
add security context config to mariadb-isalive and postgresql-isready…

raynay-r:
# Pull Request

## Description of the change

The SecurityContext configuration was not applied to the InitContainers before. This PR fixes this.

## Benefits

The SecurityContext is applied to all containers in the main Deployment and can therefore be started in Openshift with the correct configuration.

## Possible drawbacks

SecurityContext is the same for all containers of the main Deployment and can not be configured individually. Although, this is not a issue introduced here and should not be a problem.

## Checklist 

### DIFF
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -293,6 +293,10 @@ spec:
         {{- if .Values.mariadb.enabled }}
         - name: mariadb-isalive
           image: {{ .Values.mariadb.image.registry | default "docker.io" }}/{{ .Values.mariadb.image.repository }}:{{ .Values.mariadb.image.tag }}
+          {{- with .Values.nextcloud.mariadbInitContainerSecurityContext }}
+          securityContext:
+              {{- toYaml . | nindent 12 }}
+          {{- end }}
           env:
             - name: MYSQL_USER
               valueFrom:
@@ -311,6 +315,10 @@ spec:
         {{- else if .Values.postgresql.enabled }}
         - name: postgresql-isready
           image: {{ .Values.postgresql.image.registry | default "docker.io"  }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
+          {{- with .Values.nextcloud.postgresqlInitContainerSecurityContext }}
+          securityContext:
+              {{- toYaml . | nindent 12 }}
+          {{- end }}
           env:
             - name: POSTGRES_USER
               valueFrom:
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -293,7 +293,7 @@ spec:
         {{- if .Values.mariadb.enabled }}
         - name: mariadb-isalive
           image: {{ .Values.mariadb.image.registry | default "docker.io" }}/{{ .Values.mariadb.image.repository }}:{{ .Values.mariadb.image.tag }}
-          {{- with .Values.nextcloud.mariadbInitContainerSecurityContext }}
+          {{- with .Values.nextcloud.mariaDbInitContainer.securityContext }}
           securityContext:
             {{- toYaml . | nindent 12 }}
           {{- end }}
@@ -315,7 +315,7 @@ spec:
         {{- else if .Values.postgresql.enabled }}
         - name: postgresql-isready
           image: {{ .Values.postgresql.image.registry | default "docker.io"  }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
-          {{- with .Values.nextcloud.postgresqlInitContainerSecurityContext }}
+          {{- with .Values.nextcloud.postgreSqlInitContainer.securityContext }}
           securityContext:
             {{- toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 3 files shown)


provokateurin (approved):

raynay-r:
@jessebot Sorry, I completely forgot about this PR. I also think that it is not necessary to add extra parameters for the init container, but I could create them if you like.
The conflict is only with the version number. I think that's best fixed before merging, otherwise the conflict will appear again when another branch is merged earlier.

raynay-r:
I finally got around to adding the extra parameters. 

Would be great if you could take a look and sorry for the long delay (again).

raynay-r:
@jessebot the version update is done

charts/nextcloud/templates/deployment.yaml
```diff
@@ -289,6 +289,12 @@ spec:
         {{- if .Values.mariadb.enabled }}
         - name: mariadb-isalive
           image: {{ .Values.mariadb.image.registry | default "docker.io" }}/{{ .Values.mariadb.image.repository }}:{{ .Values.mariadb.image.tag }}
+          {{- if .Values.nextcloud.mariadbInitContainerSecurityContext }}
+          securityContext:
+            {{- with .Values.nextcloud.mariadbInitContainerSecurityContext }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- end }}
```
jessebot, line 297:
```suggestion
          {{- with .Values.nextcloud.mariadbInitContainerSecurityContext }}
          securityContext:
              {{- toYaml . | nindent 12 }}
          {{- end }}
```

The `with` keyword is an implicit `if`, so this makes it less wordy.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -307,6 +313,12 @@ spec:
         {{- else if .Values.postgresql.enabled }}
         - name: postgresql-isready
           image: {{ .Values.postgresql.image.registry | default "docker.io"  }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
+          {{- if .Values.nextcloud.postgresqlInitContainerSecurityContext }}
+          securityContext:
+            {{- with .Values.nextcloud.postgresqlInitContainerSecurityContext }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- end }}
```
jessebot, line 321:
```suggestion
          {{- with .Values.nextcloud.postgresqlInitContainerSecurityContext }}
          securityContext:
              {{- toYaml . | nindent 12 }}
          {{- end }}
```

The `with` keyword is an implicit `if`, so this makes it less wordy.

raynay-r:
Wow. I tried to sign the commits by rebasing and fucked up royally :D

Will try to fix it in the next few days...

raynay-r:
Ok, I recoverd the old state. The only way I know to sign these commits would be an interactive rebase, but that makes problems since there are merge commits in between. 

The only solution that comes to mind is to squash all commits and sign the resulting one.

Do you have a better idea?

provokateurin:
Squashing everything SGTM

raynay-r:
Squashed and signed. I also incremented the chart version to the next minor version.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -293,6 +293,10 @@ spec:
         {{- if .Values.mariadb.enabled }}
         - name: mariadb-isalive
           image: {{ .Values.mariadb.image.registry | default "docker.io" }}/{{ .Values.mariadb.image.repository }}:{{ .Values.mariadb.image.tag }}
+          {{- with .Values.nextcloud.mariadbInitContainerSecurityContext }}
+          securityContext:
+              {{- toYaml . | nindent 12 }}
```
provokateurin, line 298:
```suggestion
            {{- toYaml . | nindent 12 }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -311,6 +315,10 @@ spec:
         {{- else if .Values.postgresql.enabled }}
         - name: postgresql-isready
           image: {{ .Values.postgresql.image.registry | default "docker.io"  }}/{{ .Values.postgresql.image.repository }}:{{ .Values.postgresql.image.tag }}
+          {{- with .Values.nextcloud.postgresqlInitContainerSecurityContext }}
+          securityContext:
+              {{- toYaml . | nindent 12 }}
```
provokateurin, line 320:
```suggestion
            {{- toYaml . | nindent 12 }}
```

provokateurin (approved):
LGTM, just some whitespace changes to be in line with the rest of the indentations.

provokateurin:
I just thought about this again and I think it might make sense to convert these two key-value pairs into proper maps, such that they can be extended later if there are more options to set for the init-containers:
```yaml
nextcloud:
  mariaDbInitContainer:
    securityContext: {}
    ... Possible future settings
```
@raynay-r could you adapt the PR to include these changes? This makes future maintenance and additional features a bit easier.

provokateurin (approved):

provokateurin:
You need to fix the DCO and while you're at it you can squash everything again.

raynay-r:
> I just thought about this again and I think it might make sense to convert these two key-value pairs into proper maps, such that they can be extended later if there are more options to set for the init-containers:
> 
> ```yaml
> nextcloud:
>   mariaDbInitContainer:
>     securityContext: {}
>     ... Possible future settings
> ```
> 
> @raynay-r could you adapt the PR to include these changes? This makes future maintenance and additional features a bit easier.

I applied you recommendation. The only issue I see with it in the moment is that if the for example `mariaDbInitContainer` is null or the `securityContext` value is not available, the templating step will crash. Since all of that is set in the `values.yaml` it should be fine, but if you want additional null checks, let me know.

provokateurin:
I think that is fine, if you don't break it it's not broken :woman_shrugging:

PR #407, opened 2023-06-13
Update to 27.0.0

provokateurin:
# Pull Request

## Description of the change

Latest NC version

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.13
-appVersion: 26.0.2
+version: 3.5.14
+appVersion: 27.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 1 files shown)


provokateurin:
CI seems to fail because of our new Github runners.

provokateurin:
You don't need to deploy the latest helm chart if you don't want 27. You can also just override the image version instead

AndyScherzinger (approved):

provokateurin:
@jessebot Your point makes sense, although I would think that semver would always apply to the the chart and not the deployed software. You can use the latest chart and override the image with an old version and it would still work. The update only changes the default version and doesn't introduce breaking changes itself.

To get rid of this problem altogether we could make specifying the image version mandatory. Then we also don't need any updates for new versions. The docker image itself should be already tested by the people who release it. This way we really only care about the chart itself and nothing else. I'm not really a fan of doing this, but it is a possibility.

provokateurin:
Sounds good, I was just playing devil's advocate :D

PR #409, opened 2023-07-07
ServiceMonitor: Specify monitored namespace

### LINKED ISSUE #312: Feature Request: Add ServiceMonitor namespace selector
This is a small feature request that should also fix a bug in my environment.

If you install with the HelmChart the ServiceMonitor doesn't find its target as it doesn't watch any namespaces. Thus you need to patch the ServiceMonitor with the following things to let it watch the NC instance.

```
namespaceSelector:
    any: true
```

After I added that to the spec my NC was immediately scraped.

SchoolGuy:
# Pull Request

## Description of the change

This Pull Request should utilize the Helm Release Namespace as a target for the `ServiceMonitor` this Helmchart is optionally deploying.

## Benefits

The metrics actually show up in Prometheus.

## Possible drawbacks

I don't see any.

## Applicable issues

- fixes #312 

## Additional information

None

## Checklist 

- [x] ~~(optional) Variables are documented in the README.md~~

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.16
+version: 3.5.17
 appVersion: 27.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/templates/metrics-servicemonitor.yaml
+++ b/charts/nextcloud/templates/metrics-servicemonitor.yaml
@@ -24,6 +24,9 @@ spec:
       app.kubernetes.io/name: {{ include "nextcloud.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
       app.kubernetes.io/component: metrics
+  namespaceSelector:
+    matchNames:
+      - {{ .Release.Namespace | quote }}
   endpoints:
     - port: metrics
       path: "/"


SchoolGuy:
This PR should be ready for review. I never worked with Helm templates. The desired value that I want to put in there is the namespace Nextcloud is deployed into, I hope I used the right expression for that.

SchoolGuy:
@jessebot Sadly not yet. If this is a desired test, I can try to do so tomorrow.

SchoolGuy:
@jessebot If you want I can also add a PR with k3s in GitHub Actions. Or at least I can try... Then we could execute and run the HelmChart here in the repo...

SchoolGuy:
So you of course need to trust me but I did the following locally:

`values-override.yml`:

```yml
metrics:
  enabled: true
  https: true
  resources:
    requests:
      cpu: 100m
      memory: 256Mi
    limits:
      memory: 512Mi
  service:
    type: ClusterIP
    annotations:
      prometheus.io/scrape: "true"
      prometheus.io/port: "9205"
    labels: {}
  serviceMonitor:
    enabled: true
    namespace: "monitoring"
    jobLabel: "nextcloud"
    labels:
      release: kube-prom-stack
```

Output:

```
┬─[enno@tower:~/S/O/n/c/nextcloud]─[11:31:55]─[G:feature/service-monitor-namespace-selector=]
╰─>$ helm template . -n nextcloud-4seul -f values-override.yaml
---
...
---
# Source: nextcloud/templates/metrics-servicemonitor.yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: release-name-nextcloud
  namespace: "monitoring"
  labels:
    app.kubernetes.io/name: nextcloud
    helm.sh/chart: nextcloud-3.5.17
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/component: metrics
    release: kube-prom-stack
spec:
  jobLabel: "nextcloud"
  selector:
    matchLabels:
      app.kubernetes.io/name: nextcloud
      app.kubernetes.io/instance: release-name
      app.kubernetes.io/component: metrics
  namespaceSelector:
    matchNames:
      - "nextcloud-4seul"
  endpoints:
    - port: metrics
      path: "/"
      interval: 30s
```

jessebot (approved):
Looks good to me. :)

It might makes sense to make this configurable down the line if you'd like to either modify this PR or submit a second one to add a parameter to values.yaml and update the README to reflect it, but if we do that, we need a conditional in the helm template to default to the release namespace if the value is not provided.

SchoolGuy:
@jessebot What would then the content of the setting be? If we have multiple deployments of this HelmChart in the same cluster we want to either:

- Skip deployment of the `ServiceMonitor` since one is responsible for all instances.
- Ship one `ServiceMonitor` per installation (I guess the desired one) in which case I can't find the case where we would want to change this.

But my proposal would be that you open an issue for this and we discuss the details in there. :)

PR #419, opened 2023-07-21
Update nextcloud app version to `27.0.1`

jessebot:
# Pull Request

## Description of the change
Updates the default image tag to `27.0.1`

## Benefits

latest stable version as of yesterday

## Possible drawbacks / Considerations

Still no upgrade guide. This'll be the first time I'm upgrading a production instance, so happy to write up a guide on it. Is it just setting `nextcloud.update` to `1`? 🤔 Does that enable the auto-updater? Right now I have it set to the default, `0` and when I go to https://cloud.mydomain.com/updater/ it gives a 404, so maybe that's why? If I set it to `1` before upgrading the helm chart version, and just run a `helm upgrade` with the new values.yaml, will that break anything? I'll obviously take backups before this, but it would be nice to have a section of the README for this as upgrades are a common question we get in the issues.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.19
-appVersion: 27.0.0
+version: 3.5.20
+appVersion: 27.0.1
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
   - nextcloud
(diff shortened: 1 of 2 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 3.5.19
-appVersion: 27.0.0
+version: 3.5.20
```
```suggestion
version: 3.6.1
```
This can also be `3.6.1` depending on if #420 get merged before this one or not.

charts/nextcloud/README.md
```diff
@@ -28,6 +28,7 @@ helm install my-release nextcloud/nextcloud
 * [Running `occ` commands](#running-occ-commands)
     * [Putting Nextcloud into maintanence mode](#putting-nextcloud-into-maintanence-mode)
     * [Downloading models for recognize](#downloading-models-for-recognize)
+* [Upgrades](#upgrades)
```
jessebot, line 31:
```suggestion
* [Backups](#backups)
* [Upgrades](#upgrades)
```
Can add backups link if we want to keep the backups section?

tvories (approved):

PR #421, opened 2023-07-29
Use sh instead of bash as container shell in examples

provokateurin:
# Pull Request

## Description of the change

Uses sh instead of bash so the examples work in alpine based containers as well. Got introduced in https://github.com/nextcloud/helm/pull/419

## Checklist 


jessebot (approved):
thank you and sorry about missing that before 🙏

PR #440, opened 2023-09-03
adopt netxcloud.containerport in FPM deployment, allow rootless nginx

poggenpower:
# Pull Request

## Description of the change

is `netxcloud.containerport` is set, it will also configure the port for nginx. This allows nginx running in a rootless container e.g. by using their official `nginxinc/nginx-unprivileged` image. pid is written to /tmp for that reason.
Some enhanced configs parameter for service.

## Benefits
rootless nginx conaiiner. 

## Possible drawbacks

the nginx port will switch to `netxcloud.containerport` in FPM installations, but this should be
1. very rare, because `netxcloud.containerport` should not be changed 
2. even if changed the service will automatically connect to the changed port.

## Applicable issues
N/A

## Checklist 

- [x ] (optional) Variables are documented in the README.md

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -80,120 +80,122 @@ The command removes all the Kubernetes components associated with the chart and 
 
 The following table lists the configurable parameters of the nextcloud chart and their default values.
 
-| Parameter                                                  | Description                                                                            | Default                    |
-|------------------------------------------------------------|----------------------------------------------------------------------------------------|----------------------------|
-| `image.repository`                                         | nextcloud Image name                                                                   | `nextcloud`                |
-| `image.flavor`                                             | nextcloud Image type (Options: apache, fpm)                                            | `apache`                   |
-| `image.tag`                                                | nextcloud Image tag                                                                    | `appVersion`               |
-| `image.pullPolicy`                                         | Image pull policy                                                                      | `IfNotPresent`             |
-| `image.pullSecrets`                                        | Specify image pull secrets                                                             | `nil`                      |
-| `replicaCount`                                             | Number of nextcloud pods to deploy                                                     | `1`                        |
-| `ingress.className`                                        | Name of the ingress class to use                                                       | `nil`                      |
-| `ingress.enabled`                                          | Enable use of ingress controllers                                                      | `false`                    |
-| `ingress.servicePort`                                      | Ingress' backend servicePort                                                           | `http`                     |
-| `ingress.annotations`                                      | An array of service annotations                                                        | `nil`                      |
-| `ingress.labels`                                           | An array of service labels                                                             | `nil`                      |
-| `ingress.path`                                             | The `Path` to use in Ingress' `paths`                                                  | `/`                        |
-| `ingress.pathType`                                         | The `PathType` to use in Ingress' `paths`                                              | `Prefix`                   |
-| `ingress.tls`                                              | Ingress TLS configuration                                                              | `[]`                       |
-| `nextcloud.host`                                           | nextcloud host to create application URLs, updates trusted_domains at installation time only| `nextcloud.kube.home` |
-| `nextcloud.username`                                       | User of the application                                                                | `admin`                    |
-| `nextcloud.password`                                       | Application password                                                                   | `changeme`                 |
-| `nextcloud.existingSecret.enabled`                         | Whether to use an existing secret or not                                               | `false`                    |
-| `nextcloud.existingSecret.secretName`                      | Name of the existing secret                                                            | `nil`                      |
-| `nextcloud.existingSecret.usernameKey`                     | Name of the key that contains the username                                             | `nil`                      |
-| `nextcloud.existingSecret.passwordKey`                     | Name of the key that contains the password                                             | `nil`                      |
-| `nextcloud.existingSecret.smtpUsernameKey`                 | Name of the key that contains the SMTP username                                        | `nil`                      |
-| `nextcloud.existingSecret.smtpPasswordKey`                 | Name of the key that contains the SMTP password                                        | `nil`                      |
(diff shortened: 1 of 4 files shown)


charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -44,7 +32,7 @@ data:
         }
 
         server {
-            listen 80;
+            listen {{ .Values.nextcloud.containerPort }};
```
jessebot, line 35: Would it make more sense to allow specifying an `nginx.containerPort` parameter instead of defaulting using the existing nextcloud port?

charts/nextcloud/README.md
```diff
@@ -96,7 +96,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `ingress.path`                                             | The `Path` to use in Ingress' `paths`                                                  | `/`                        |
 | `ingress.pathType`                                         | The `PathType` to use in Ingress' `paths`                                              | `Prefix`                   |
 | `ingress.tls`                                              | Ingress TLS configuration                                                              | `[]`                       |
-| `nextcloud.host`                                           | nextcloud host to create application URLs                                              | `nextcloud.kube.home`      |
+| `nextcloud.host`                                           | nextcloud host to create application URLs, updates trusted_domains at installation time only| `nextcloud.kube.home` |
```
jessebot, line 99: wait, is this the case?

https://github.com/nextcloud/helm/blob/268defe1fff1af02a83fe2f1f9d22e9a71e0f797/charts/nextcloud/templates/_helpers.tpl#L180-L181

It looks like it's specified as an env var and according to the nextcloud/docker repo:
https://github.com/nextcloud/docker/blob/f9ae675c1ac2aed735435e84dd1794eb28890103/README.md#L138-L138

I think it's done after installation. Are you saying it has no effect if set on an already running helm release?

jessebot (dismissed):
overall seems like a good change, but have a few comments to address

charts/nextcloud/README.md
poggenpower: Yes it seems this is a limitation of the container. At least it was the behavior it monitored.

charts/nextcloud/templates/nginx-config.yaml
poggenpower: Have thought about it. But at the end it is still the port how nextcloud is accessible on container level. Migrating from non FPM to FPM would continue to work in the same manner. At the end in the non FPM case it is the port of the apache. But if `nextcould.` is more pointing to the settings the affect the container running the nextcloud image, then moving to `nginx.containerport` would make sense.
I have no strong opinion, so please let me know what you prefer.

charts/nextcloud/README.md
provokateurin: We have other variables that behave the same way, for example the database settings. Those settings are only copied at install time which is obviously pretty bad in a kubernetes environment. I wanted to open a PR for that but haven't found the time yet.

poggenpower: @provokateurin  Yes, the way the nextcloud docker container works is more than ideal. Its logic is more made from "maintain nextclould on a server" prosective in mind, instead of an idempotent container. e.g. installing nextcloud instead of having it already in place. 
Futher i would like to see a more modular entrypoint, that allows to write all config in a separate step. This would allow to use an init container with root privileges to e.g. write the php-redis config and then run the main container without root.

provokateurin: There is `nextcloud.configs` which can be used to write custom configs. Those would be evaluated at chart install time. This is the way to properly implement it. I think this discussion should happen in a new issue as it is not directly related to this PR and also concerns more of the configs than just the trusted domains.

charts/nextcloud/README.md
```diff
@@ -80,118 +80,122 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `nginx.enabled`                                            | Enable nginx (requires you use php-fpm image)                                                | `false`                    |
+| `nginx.image.repository`                                   | nginx Image name, e.g. use `nginxinc/nginx-unprivileged` for rootless container              | `nginx`                    |
+| `nginx.image.tag`                                          | nginx Image tag                                                                              | `alpine`                   |
+| `nginx.image.pullPolicy`                                   | nginx Image pull policy                                                                      | `IfNotPresent`             |
+| `nginx.image.pullPolicy`                                   | nginx Image pull policy                                                                      | `IfNotPresent`             |
+| `nginx.image.containerPort`                                | Customize container port e.g. when not running as root                                       | `IfNotPresent`             |
```
jessebot, line 144:
```suggestion
| `nginx.containerPort`                                      | Customize container port e.g. when not running as root                                       | `IfNotPresent`             |
```

charts/nextcloud/templates/service.yaml
```diff
@@ -13,6 +13,13 @@ spec:
...
+  {{- if .Values.service.ipFamilies }}
+  ipFamilies: 
+  {{- toYaml .Values.service.ipFamilies | nindent 4 }}
+  {{- end }}
+  {{- if .Values.service.ipFamilyPolicy }}
+  ipFamilyPolicy: {{ .Values.service.ipFamilyPolicy }}
+  {{- end }}
   ports:
   - port: {{ .Values.service.port }}
     targetPort: {{ .Values.nextcloud.containerPort }}
```
jessebot, line 25: Should this have a conditional to be the nginx port if nginx.containerPort is set?

charts/nextcloud/README.md
jessebot: Can you also please add the parameter to values.yaml?

charts/nextcloud/templates/service.yaml
poggenpower: No, this is independent from the container port and a service setting. Mainly needed if want to make your service reachable via IPv6

poggenpower:
@jessebot  wasn't able to resolve this too, I have created a new PR #444 to workaround.

PR #442, opened 2023-09-05
updating postgres-is-ready init container to use the same hostname env var template as the nextcloud container

jessebot:
# Pull Request

## Description of the change

This changes the template for the `postgresql-is-ready` init container to use the same hostname variable, `POSTGRES_HOST` that we use for the actual nextcloud container.

## Benefits

This should make it so that if you use `postgresql.enabled=true` and `postgresql.fullnameOverride=my-postgres-name` that we use that for checking if postgresql is ready as well. Previously, we were using `{.Release.name}-postgresql` which would not accommodate `postgresql.fullnameOverride`.

## Possible drawbacks

None that I can think of, but as always, open to feedback. If there's a better hostname to use instead of the primary hostname, I'm happy to change to that though I only saw a few services in the [bitnami postgresql templates](https://github.com/bitnami/charts/tree/main/bitnami/postgresql/templates) and those were the primary, primary headless, and read replicas. Read could work, but I actually don't know what primary headless does?

## Checklist 


provokateurin (approved):

charts/nextcloud/Chart.yaml
```diff
@@ -33,4 +33,4 @@ dependencies:
   - name: redis
     version: 17.13.*
     repository: oci://registry-1.docker.io/bitnamicharts
-    condition: redis.enabled
+    condition: redis.enabled
```
jessebot, line 36:
```suggestion
    condition: redis.enabled

```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -298,10 +298,12 @@ spec:
               secretKeyRef:
                 name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-%s" .Release.Name "db") }}
                 key: {{ .Values.externalDatabase.existingSecret.usernameKey | default "db-username" }}
+          - name: POSTGRES_HOST
+            value: {{ template "postgresql.primary.fullname" .Subcharts.postgresql }}
         command:
           - "sh"
           - "-c"
-          - {{ printf "until pg_isready -h %s-postgresql -U ${POSTGRES_USER} ; do sleep 2 ; done" .Release.Name }}
+          - {{ printf "until pg_isready -h ${POSTGRES_HOST} -U ${POSTGRES_USER} ; do sleep 2 ; done" }}
```
jessebot, line 306:
```suggestion
          - "until pg_isready -h ${POSTGRES_HOST} -U ${POSTGRES_USER} ; do sleep 2 ; done"
```
I'm not sure we even need the printf anymore, because we're using exclusively env vars

provokateurin: Should work :)

PR #444, opened 2023-09-05
 adopt netxcloud.containerport in FPM deployment, allow rootless nginx - nd2 try

poggenpower:
replaces #440 
as the old PR got stuck in git fanciness, simply resubmit as a new PR.
Seems to be the easiest way

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -80,120 +80,122 @@ The command removes all the Kubernetes components associated with the chart and 
 
 The following table lists the configurable parameters of the nextcloud chart and their default values.
 
-| Parameter                                                  | Description                                                                            | Default                    |
-|------------------------------------------------------------|----------------------------------------------------------------------------------------|----------------------------|
-| `image.repository`                                         | nextcloud Image name                                                                   | `nextcloud`                |
-| `image.flavor`                                             | nextcloud Image type (Options: apache, fpm)                                            | `apache`                   |
-| `image.tag`                                                | nextcloud Image tag                                                                    | `appVersion`               |
-| `image.pullPolicy`                                         | Image pull policy                                                                      | `IfNotPresent`             |
-| `image.pullSecrets`                                        | Specify image pull secrets                                                             | `nil`                      |
-| `replicaCount`                                             | Number of nextcloud pods to deploy                                                     | `1`                        |
-| `ingress.className`                                        | Name of the ingress class to use                                                       | `nil`                      |
-| `ingress.enabled`                                          | Enable use of ingress controllers                                                      | `false`                    |
-| `ingress.servicePort`                                      | Ingress' backend servicePort                                                           | `http`                     |
-| `ingress.annotations`                                      | An array of service annotations                                                        | `nil`                      |
-| `ingress.labels`                                           | An array of service labels                                                             | `nil`                      |
-| `ingress.path`                                             | The `Path` to use in Ingress' `paths`                                                  | `/`                        |
-| `ingress.pathType`                                         | The `PathType` to use in Ingress' `paths`                                              | `Prefix`                   |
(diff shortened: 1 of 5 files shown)


charts/nextcloud/README.md
```diff
@@ -80,118 +80,122 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `nginx.enabled`                                            | Enable nginx (requires you use php-fpm image)                                                | `false`                    |
+| `nginx.image.repository`                                   | nginx Image name, e.g. use `nginxinc/nginx-unprivileged` for rootless container              | `nginx`                    |
+| `nginx.image.tag`                                          | nginx Image tag                                                                              | `alpine`                   |
+| `nginx.image.pullPolicy`                                   | nginx Image pull policy                                                                      | `IfNotPresent`             |
+| `nginx.image.pullPolicy`                                   | nginx Image pull policy                                                                      | `IfNotPresent`             |
+| `nginx.image.containerPort`                                | Customize container port e.g. when not running as root                                       | `IfNotPresent`             |
```
jessebot, line 144:
```suggestion
| `nginx.containerPort`                                     | Customize container port e.g. when not running as root                                       | `IfNotPresent`             |
```

jessebot (changes requested):
Please update the values.yaml and the one other comment I left and you should be good to go :)

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -1,20 +1,8 @@
...
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
-data:
-  nginx.conf: |-
-{{- if .Values.nginx.config.default }}
+{{- define "nginx.conf" }}
     worker_processes auto;
 
     error_log  /var/log/nginx/error.log warn;
-    pid        /var/run/nginx.pid;
+    pid        /tmp/nginx.pid;
```
jessebot, line 5: will this break any normal flows not using a non-root container?

jessebot (approved):

charts/nextcloud/templates/nginx-config.yaml
poggenpower: This pid stuff is not really required in containers, as the container get removed if something goes wrong, to pid to check and kill afterwards etc. As root has also permission to /tmp this change doesn't harm.

charts/nextcloud/templates/nginx-config.yaml
```diff
@@ -44,7 +32,7 @@ data:
         }
 
         server {
-            listen 80;
+            listen {{ .Values.nginx.containerPort | default "80" }};
```
tilgovi, line 35: Sorry to miss this before it merged, but I'm curious why not just `.Values.nextcloud.containerPort`? I'm not seeing a reason why a new variable was necessary, and now it seems like it's necessary to set both variables. The `deployment.yaml` template still uses `.Values.nextcloud.containerPort` for the `containerPort` of the nginx container.

poggenpower: This was the case in my original PR, but we discussed this here https://github.com/nextcloud/helm/pull/440#discussion_r1314479033 and ended up with an extra var.

jessebot: we could change the default to be containerPort, so the new line would be like:

```
listen {{ .Values.nginx.containerPort | default .Values.nextcloud.containerPort }};
```

Would that help?

tilgovi: Thanks for pointing out the previous discussion. I don't follow the argument for granularity here, but if that feels right to you both then we should keep it.

The template update to fall back to the primary container port variable seems convenient, but I'm fine to leave things as they are.

It's only because I didn't understand any current or future scenarios where these variables would *ever* have different values, and didn't see the previous discussion, that I suggested just removing the nginx one.

PR #461, opened 2023-11-05
[cron] run a system cron job inside the nextcloud image

MichaelSp:
# Pull Request

## Description of the change

Instead of running the cron as a sidecar, with this change the cron can run inside the main container by setting `cronjob.asSidecar=false`. Default will stay `true` for backwards compatability.

## Benefits

run cron inside main container

## Possible drawbacks

none. Backwards comparability maintained

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.3.6
+version: 4.3.7
 appVersion: 27.1.3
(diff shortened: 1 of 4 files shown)


provokateurin:
Why do you want to have this? I don't see any benefit in this other than that it could break when messing with the command line.

MichaelSp:
hi @provokateurin 

I tried to work around a problem I had with the sidecar cron. Turns out the cron works just fine. Can't recall what the issue was. 
Looks like we can close this.

PR #462, opened 2023-11-05
helm unittests

MichaelSp:
# Pull Request

## Description of the change

run unittests for the helm-chart during PR checks

## Benefits

better quality

## Possible drawbacks

??

## Applicable issues

no

## Checklist 


provokateurin:
Please explain what this does exactly. I'm not familiar with helm-unittest and don't even know what the tool is supposed to do.

MichaelSp:
https://github.com/helm-unittest/helm-unittest is a tool to help write unit tests for helm charts. It allows to write unit tests which help ensure software quality on the unit level.

Here is a very good article on what unit-tests actually are: https://martinfowler.com/bliki/UnitTest.html

provokateurin:
I know what unit tests are, but what is tested in the context of Helm Charts?

MichaelSp:
In this https://github.com/nextcloud/helm/blob/2c871d9bdfbddc748cee7a0aa575fbc32a04b0f2/charts/nextcloud/tests/defaults_test.yaml I test the output of the helm-chart for a default value file. I only provide values to 
 * `Chart.Version` - https://github.com/nextcloud/helm/blob/2c871d9bdfbddc748cee7a0aa575fbc32a04b0f2/charts/nextcloud/tests/defaults_test.yaml#L7C3-L7C10
 * `Chart.appVersion` - https://github.com/nextcloud/helm/blob/2c871d9bdfbddc748cee7a0aa575fbc32a04b0f2/charts/nextcloud/tests/defaults_test.yaml#L8
 * and `metrics.token` - https://github.com/nextcloud/helm/blob/2c871d9bdfbddc748cee7a0aa575fbc32a04b0f2/charts/nextcloud/tests/defaults_test.yaml#L12

This is necessary to generate stable output (it should not change between test runs even if the `appVersion` changes.
The output is matched and compared to [the snapshot ](https://github.com/nextcloud/helm/blob/2c871d9bdfbddc748cee7a0aa575fbc32a04b0f2/charts/nextcloud/tests/__snapshot__/defaults_test.yaml.snap) to make sure future changes to the templates or values don't change the output in an unexpected way.

If a future change results in a desired change of the output, the snapshot can be updated: `helm unittest charts/nextcloud -u`.

This is only a very basic test. You can increase coverage by defining more tests and more test-suites with different combinations of `values` to cover the logic in the templates and make sure use-cases like nginx, cron, mail, etc result in the correct configuration.

Using `helm unittest` is a best-practice in helm-charts.

provokateurin:
Sounds like a good idea, @jessebot @tvories what do you think? I think for merging this it would be good to have a bit more coverage of features

charts/nextcloud/tests/__snapshot__/defaults_test.yaml.snap
```diff
@@ -0,0 +1,145 @@
...
+          app.kubernetes.io/component: app
+          app.kubernetes.io/instance: RELEASE-NAME
+          app.kubernetes.io/name: nextcloud
+      strategy:
+        type: Recreate
+      template:
+        metadata:
+          annotations:
+            nextcloud-config-hash: a5aae02b1b8278a9c8a2dc143e82d3737fc295f62c34afd617207f37d1b2b438
+            php-config-hash: 44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a
```
jessebot, line 26: would we need to keep these up to date each version of the nextcloud docker container we update?

jessebot: nvm, I think this part answers my question:
> If a future change results in a desired change of the output, the snapshot can be updated: helm unittest charts/nextcloud -u.

jessebot (approved):

MichaelSp:
> there's also the question of how we run only necessary unit tests after each push? Otherwise, every push will take like 5+ minutes as we add more feature unit tests, which will also make the back and forth on PRs take a bit longer.

The unit-tests even for larger test suites are pretty fast (<5sec). I bet they will be always faster than starting a kind-cluster and deploy it.

provokateurin:
I assume the unit tests just run `helm template` with a given values file and compare the output with the fixture? Then that would be very fast and scalable indeed.

provokateurin:
@jessebot @MichaelSp what is the status? I think it would be good to start adding tests soon, even if there are just a few in the beginning.

provokateurin:
I'd say the `charts/__snapshots__` dir needs to be added to the filter so that the CI runs when they change to ensure everything is still correct.

.github/workflows/lint-test.yaml
```diff
@@ -20,6 +18,7 @@ jobs:
               - 'charts/nextcloud/Chart.yaml'
               - 'charts/nextcloud/values.yaml'
               - 'charts/nextcloud/templates/**'
+              - 'charts/nextcloud/tests/**'
```
jessebot, line 21:
```suggestion
```
@provokateurin We'd need to either A) remove this to make the test pass since it wants to run the lint (then submit a new PR for adding this to the filter), or B) bump the `version` in Chart.yaml a patch version to make the linter happy.

jessebot (commented):
I regenerated the snapshot for today and rebased off of main, but I have one last comment below to take care of before we can merge this.

.github/workflows/lint-test.yaml
provokateurin: I can also just merge with red CI, that would be fine for me in this case as it makes no sense to bump the version and having another PR is also not that useful.

jessebot: yeah, I think just merging iwth red CI is fine, since as you said, it makes no sense to bump the version in this case.

provokateurin:
I think this might still be a good idea to prevent any accidental issues with the templating that otherwise might go undetected.

MichaelSp:
Finally I found the time to come back to this and:

 * add the requested hint to CONTRIBUTING.md
 * Test case 1: [a valid helm template change](https://github.com/MichaelSp/nextcloud-helm/pull/49)
 * Test case 2: [invalid helm template change](https://github.com/MichaelSp/nextcloud-helm/pull/50)

volker-raschek:
Hello @provokateurin , 
I have experienced frequent recursions in my cluster regarding the Kubernetes helm chart. 

Just recently I wrote a helm chart for the [prometheus-postgres-exporter](https://git.cryptic.systems/volker.raschek/prometheus-postgres-exporter) which covers some unit tests with the helm plugin [unittest](https://github.com/helm-unittest/helm-unittest). Furthermore, it also includes linters, a generated README and other good tooling. The plugin is also used on other projects like [gitea](https://gitea.com/gitea/helm-chart), where I recently opened a [PR](https://gitea.com/gitea/helm-chart/pulls/772).

For this reason I can only recommend the endeavour of @MichaelSp  and wonder why the plugin `unittest` was not looked at more closely a year ago, to avoid recursion.

MichaelSp:
anything missing @provokateurin @jessebot  @tvories ?

MichaelSp:
Thanks for you approval @jessebot 
Anything else missing @tvories @provokateurin ?

MichaelSp:
Hey @wrenix can you pls have a look at this PR? AFAIK there is only a review missing.

MichaelSp:
re-based again. Please check and approve

MichaelSp:
@wrenix  let me know if you want me to rebase again....

volker-raschek:
Hi @MichaelSp, 
can you please rebase your PR? The logs are no longer available.

Alternatively, can you or @wrenix retrigger the CI pipeline?

volker-raschek (approved):

PR #463, opened 2023-11-06
Update issue.md - add note about syntax highlighting

jessebot:
# Pull Request

## Description of the change

Adds polite note about keeping the <code>```yaml</code> in place, because I can't read so great and it _really_ helps me.

## Benefits

Maybe some users remove this by accident or because they think it is just left over by accident and this will help them.

## Possible drawbacks

slightly longer issue template by one line :P


provokateurin (approved):
Good idea :)

PR #465, opened 2023-11-11
chore: lint all templates to fix indentation and update style

wrenix:
# Pull Request

## Description of the change

- replace indent with nindent (and spaces)
- array of yaml indent
- with+if replace with with (with is already an if)
- nginx if over all probes
- use more with
- default "just" in `values.yaml`

## Benefits

readable code

## Possible drawbacks

none

## Applicable issues
none

## Additional information

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.0
+version: 4.5.1
 appVersion: 27.1.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/templates/db-secret.yaml
+++ b/charts/nextcloud/templates/db-secret.yaml
@@ -10,16 +10,16 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 type: Opaque
-stringData:
+data:
   {{- if .Values.mariadb.enabled }}
-  db-username: {{ .Values.mariadb.auth.username | quote }}
-  db-password: {{ .Values.mariadb.auth.password | quote }}
+  db-username: {{ .Values.mariadb.auth.username | b64enc | quote }}
+  db-password: {{ .Values.mariadb.auth.password | b64enc | quote }}
   {{- else if .Values.postgresql.enabled }}
-  db-username: {{ .Values.postgresql.global.postgresql.auth.username | quote }}
-  db-password: {{ .Values.postgresql.global.postgresql.auth.password | quote }}
+  db-username: {{ .Values.postgresql.global.postgresql.auth.username | b64enc | quote }}
+  db-password: {{ .Values.postgresql.global.postgresql.auth.password | b64enc | quote }}
   {{- else }}
-  db-username: {{ .Values.externalDatabase.user | quote }}
-  db-password: {{ .Values.externalDatabase.password | quote }}
+  db-username: {{ .Values.externalDatabase.user | b64enc | quote }}
+  db-password: {{ .Values.externalDatabase.password | b64enc | quote }}
   {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/nextcloud/templates/db-secret.yaml
+++ b/charts/nextcloud/templates/db-secret.yaml
@@ -3,23 +3,23 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ printf "%s-%s" .Release.Name "db"  }}
+  name: {{ .Release.Name }}-db
   labels:
     app.kubernetes.io/name: {{ include "nextcloud.name" . }}
     helm.sh/chart: {{ include "nextcloud.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 type: Opaque
-data:
+stringData:
   {{- if .Values.mariadb.enabled }}
-  db-password: {{ default "" .Values.mariadb.auth.password | b64enc | quote }}
-  db-username: {{ default "" .Values.mariadb.auth.username | b64enc | quote }}
+  db-username: {{ .Values.mariadb.auth.username | quote }}
+  db-password: {{ .Values.mariadb.auth.password | quote }}
   {{- else if .Values.postgresql.enabled }}
-  db-password: {{ default "" .Values.postgresql.global.postgresql.auth.password | b64enc | quote }}
-  db-username: {{ default "" .Values.postgresql.global.postgresql.auth.username | b64enc | quote }}
+  db-username: {{ .Values.postgresql.global.postgresql.auth.username | quote }}
+  db-password: {{ .Values.postgresql.global.postgresql.auth.password | quote }}
   {{- else }}
-  db-password: {{ default "" .Values.externalDatabase.password | b64enc | quote }}
-  db-username: {{ default "" .Values.externalDatabase.user | b64enc | quote }}
+  db-username: {{ .Values.externalDatabase.user | quote }}
+  db-password: {{ .Values.externalDatabase.password | quote }}
   {{- end }}
 {{- end }}
 {{- end }}
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -59,18 +59,18 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           {{- with .Values.lifecycle }}
           lifecycle:
-            {{-   with .postStartCommand }}
+            {{- with .postStartCommand }}
             postStart:
               exec:
                 command:
                   {{- toYaml . | nindent 18 -}}
-            {{-   end }}
-            {{-   with .preStopCommand }}
+            {{- end }}
+            {{- with .preStopCommand }}
             preStop:
               exec:
                 command:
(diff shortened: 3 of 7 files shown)


charts/nextcloud/values.yaml
```diff
@@ -261,8 +261,8 @@ externalDatabase:
   existingSecret:
     enabled: false
     # secretName: nameofsecret
```
jessebot, line 263:
```suggestion
    secretName: ""
```

charts/nextcloud/templates/db-secret.yaml
```diff
@@ -3,23 +3,23 @@
...
   {{- else if .Values.postgresql.enabled }}
-  db-password: {{ default "" .Values.postgresql.global.postgresql.auth.password | b64enc | quote }}
-  db-username: {{ default "" .Values.postgresql.global.postgresql.auth.username | b64enc | quote }}
+  db-username: {{ .Values.postgresql.global.postgresql.auth.username | quote }}
+  db-password: {{ .Values.postgresql.global.postgresql.auth.password | quote }}
   {{- else }}
-  db-password: {{ default "" .Values.externalDatabase.password | b64enc | quote }}
-  db-username: {{ default "" .Values.externalDatabase.user | b64enc | quote }}
+  db-username: {{ .Values.externalDatabase.user | quote }}
+  db-password: {{ .Values.externalDatabase.password | quote }}
```
jessebot, line 22: We shouldn't have this as stringData, because we don't want any templates with plain text passwords. I know kubernetes will convert this after it's applied/created in server, but this would mean anything that inflates these templates before applying them, would have plain text passwords, even though I know base64 encoding isn't encryption anyway. Kubernetes Secrets are... weird.

Still good to remove the defaults there though. :+1:

charts/nextcloud/templates/deployment.yaml
```diff
@@ -8,17 +9,17 @@ metadata:
...
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/component: app
-  {{- if .Values.deploymentLabels }}
-{{ toYaml .Values.deploymentLabels | indent 4 }}
-  {{- end }}
-  {{- if .Values.deploymentAnnotations }}
+    {{- with .Values.deploymentLabels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.deploymentAnnotations }}
```
jessebot, line 15: We still need `if`s wrapping the `with`s, no?

charts/nextcloud/values.yaml
wrenix: sadly there is a strange/bad logic on the initContainers in `deployment.yaml`: https://github.com/nextcloud/helm/pull/465/files#diff-9320b77c358085c20797303fd29b5f51cc0a94ee926042a12ba4e0011053deaaR288

i believe this check get broken, on "per-deliverd" databases

charts/nextcloud/templates/deployment.yaml
wrenix: no `with` is an implicite `if`:
you could also use:
```yaml
{{- with .Values.donotexists }}
enabled: true
value: {{ . }}
{{- else }}
enabled: false
{{- end }}
```

jessebot (commented):
Left a few comments, but overall good linting/formatting changes :) What did you use to do this and in which text editor/IDE?

charts/nextcloud/templates/deployment.yaml
jessebot: oh that's awesome! brb updating several of my own helm charts :D 

TIL: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

wrenix: https://pkg.go.dev/text/template
> {{with pipeline}} T1 {{else}} T0 {{end}}
>	If the value of the pipeline is empty, dot is unaffected and T0
>	is executed; otherwise, dot is set to the value of the pipeline
>	and T1 is executed.

charts/nextcloud/values.yaml
wrenix: but yes, maybe we should solve it with an helper - but current i just do linting and formating here

you could create an issue for it and assign it to me

jessebot: :thinking: hmmm, I'll take a look when I have little more time, thank you for the helpful link :pray:

wrenix:
sadly i do that handly and use the https://helix-editor.com/ - but i see that not readible code very often (and start sometimes with some cleanups before i work there)

PS: i am thinking of createn an PR, to manage nextcoud-apps and there configuration with an initcontainer and script using occ-commands ...

charts/nextcloud/values.yaml
jessebot: I will try to circle back to that, but in the meantime, you can create the issue yourself and if you have time, we'd also love a PR :blue_heart:

wrenix:
so i hope for the first step, i have everything solved.

PS: I create a discussion issue for my idea of application management, see #466

jessebot (approved):
This looks good to me and I learned some stuff along the way, which is always awesome! Due to the size of this PR, I'm going to tag in a couple of other collaborators to review this and make sure I didn't miss anything.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.4.0
```
jessebot, line 3: Merged a PR before this, so this needs to be bumped in line with the new chart version, please

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
-        resources:
-{{ toYaml .Values.resources | indent 10 }}
-        {{- if .Values.nextcloud.securityContext}}
-        securityContext:
+        - name: {{ .Chart.Name }}
+          image: {{ include "nextcloud.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- with .Values.lifecycle }}
+          lifecycle:
+            {{-   with .postStartCommand }}
```
provokateurin, line 62:
```suggestion
            {{- with .postStartCommand }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            postStart:
+              exec:
+                command:
+                  {{- toYaml . | nindent 18 -}}
+            {{-   end }}
```
provokateurin, line 67:
```suggestion
            {{- end }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            {{-   with .preStopCommand }}
```
provokateurin, line 68:
```suggestion
            {{- with .preStopCommand }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            preStop:
+              exec:
+                command:
+                  {{- toYaml . | nindent 18 -}}
+            {{-   end }}
```
provokateurin, line 73:
```suggestion
            {{- end }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+        {{- end }}{{/* end-if nginx.enabled */}}
+        {{- if .Values.cronjob.enabled }}
+        - name: {{ .Chart.Name }}-cron
+          image: {{ include "nextcloud.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+          - /cron.sh
+          {{- with .Values.cronjob.lifecycle }}
+          lifecycle:
+            {{-   with .postStartCommand }}
```
provokateurin, line 245:
```suggestion
            {{- with .postStartCommand }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            postStart:
+              exec:
+                command:
+                  {{- toYaml . | nindent 18 -}}
+            {{-   end }}
```
provokateurin, line 250:
```suggestion
            {{- end }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            {{-   with .preStopCommand }}
```
provokateurin, line 251:
```suggestion
            {{- with .preStopCommand }}
```

charts/nextcloud/templates/deployment.yaml
```diff
@@ -33,337 +34,339 @@ spec:
...
+            preStop:
+              exec:
+                command:
+                  {{- toYaml . | nindent 18 -}}
+            {{-   end }}
```
provokateurin, line 256:
```suggestion
            {{- end }}
```

provokateurin (approved):
LGTM, just some whitespace nitpicks

jessebot (approved):

DanishVaid:
@wrenix I think this introduced a regression. The change in `charts/nextcloud/templates/config.yaml` line 14 from `indent` -> `nindent` causes an extra newline and that leads to a PHP error:

Config file has leading content, please remove everything before "<?php" in custom.config.php
Fatal error: Uncaught Error: Typed static property OC::$server must not be accessed before initialization in /var/www/html/index.php:71 Stack trace: #0 {main} thrown in /var/www/html/index.php on line 71

PR #472, opened 2023-11-17
Smtp documentation update

### LINKED ISSUE #236: Double domainname in smtp in email server admin settings
When I'm configuring nextcloud like this:

```
  existingSecret:
    enabled: true
    secretName: mysecret
    usernameKey: username
    passwordKey: password
    smtpUsernameKey: smtp_username
    smtpPasswordKey: smtp_password
  update: 0
  datadir: /var/www/html/data
  persistence:
    subPath:
  mail:
    enabled: true
    fromAddress: no-reply@domain.org
    domain: domain.org
    smtp:
      host: host.org
      secure: ssl
      port: 465
      authtype: LOGIN
      name: name
```

and this in the secret:

`smtp_username=no-reply@domain.org`

I find this in UI of basic settings panel of Next Cloud:

`From address no-reply@domain.org @ domain.org`

If I change this in UI it gets rewritten to original settings after reload. What am I doing wrong, how to fix this behaviour and have `domain.org` only once?

MaximUltimatum:
# Pull Request

## Description of the change

Documentation update to close out issue #236  

## Benefits

Clearer documentation to avoid a frequent trip-up with users when filling in the values.yaml file

## Possible drawbacks

Slightly wordier documentation

## Applicable issues

- fixes #236 

## Additional information

Opted to put this clarifying note in just the README and not the values file itself to avoid clutter. If someone is confused about what to put in the username, or having issues, they will naturally check the README, where this should then help them.

## Checklist 

- [X ] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [ X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X] (optional) Variables are documented in the README.md

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.1
+version: 4.5.0
 appVersion: 27.1.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.0
+version: 4.5.1
 appVersion: 27.1.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.0
+version: 4.5.1
```
```suggestion
version: 4.5.0
```
You can remove the version bump because this is only the README, not the values.yaml or templated files :)

jessebot (changes requested):
looks good, just remove the version bump only because the chart itself is not changing. Thanks for clarification :)

jessebot (approved):
just realized I had permissions to edit this, so I removed the chart version bump. thank you!

Issue #475, opened 2023-11-18
[bug] extra newline in config files 


ishioni:
## Describe your Issue
Chart version 4.5.1 replaced indent with nindent for config files. This unfortunately ends up rendering those config files with a newline prepended

### Logs and Errors
As you can see in my local [flux repo PR](https://github.com/ishioni/homelab-ops/pull/1478), extra config files are now rendered with a newline and a  |2- instead of a |-, and an added newline as a bonus. This in turns errors out nextcloud with an error

```
Config file has leading content, please remove everything before "<?php" in local.config.php
Fatal error: Uncaught Error: Typed static property OC::$server must not be accessed before initialization in /var/www/html/index.php:71 Stack trace: #0 {main} thrown in /var/www/html/index.php on line 71
```

## Describe your Environment

- Kubernetes distribution: talos v1.5.5

- Helm Version (or App that manages helm): flux v2.1.2

- Helm Chart Version: 4.5.1

- `values.yaml`:

```yaml
deploymentAnnotations:
  secret.reloader.stakater.com/reload: &secret nextcloud-secret
image:
  repository: public.ecr.aws/docker/library/nextcloud
  tag: 27.1.3-fpm-alpine
nextcloud:
  host: &host "files.${SECRET_DOMAIN}"
  extraInitContainers:
    - name: init-db
      image: ghcr.io/onedr0p/postgres-init:14.9
      envFrom:
        - secretRef:
            name: *secret
  extraEnv:
    - name: S3_ACCESS_KEY
      valueFrom:
        secretKeyRef:
          name: *secret
          key: S3_ACCESS_KEY
    - name: S3_SECRET_KEY
      valueFrom:
        secretKeyRef:
          name: *secret
          key: S3_SECRET_KEY
    - name: OIDC_CLIENT_ID
      valueFrom:
        secretKeyRef:
          name: *secret
          key: OIDC_CLIENT_ID
    - name: OIDC_CLIENT_PASSWORD
      valueFrom:
        secretKeyRef:
          name: *secret
          key: OIDC_CLIENT_PASSWORD
  existingSecret:
    enabled: true
    secretName: *secret
    usernameKey: ADMIN_USER
    passwordKey: ADMIN_PASS
    tokenKey: METRICS_TOKEN
    smtpHostKey: SMTP_HOST
    smtpUsernameKey: SMTP_USERNAME
    smtpPasswordKey: SMTP_PASSWORD
  mail:
    enabled: true
    fromAddress: files
[...]

PR #476, opened 2023-11-19
fix: after #465 (remove before <?php, indent of extraVolumes and podLabels)

wrenix:
Fixes:
  - "remove everthing before <?php":
    - https://github.com/nextcloud/helm/pull/465#issuecomment-1817669544
    - https://github.com/nextcloud/helm/issues/475
  - indent of extraVolumes: https://github.com/nextcloud/helm/issues/475#issuecomment-1817884236
  - inside of `with` using `.Values` (e.g. podLabels)

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.1
+version: 4.5.2
 appVersion: 27.1.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/templates/config.yaml
+++ b/charts/nextcloud/templates/config.yaml
@@ -11,7 +11,7 @@ metadata:
 data:
   {{- range $key, $value := .Values.nextcloud.configs }}
   {{ $key }}: |-
-    {{ $value | nindent 4 }}
+    {{- $value | nindent 4 }}
   {{- end }}
   {{- if .Values.nextcloud.defaultConfigs }}
   {{- if index .Values.nextcloud.defaultConfigs ".htaccess" }}
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -35,7 +35,7 @@ spec:
         {{ template "nextcloud.redis.fullname" . }}-client: "true"
         {{- end }}
         {{- with .Values.podLabels }}
-        {{- toYaml .Values.podLabels | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
         {{- end }}
       annotations:
         nextcloud-config-hash: {{ print (toJson .Values.nextcloud.defaultConfigs) "-" (toJson .Values.nextcloud.configs) | sha256sum }}
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -350,7 +350,7 @@ spec:
             name: {{ template "nextcloud.fullname" . }}-nginxconfig
         {{- end }}
         {{- with .Values.nextcloud.extraVolumes }}
-        {{- toYaml . | nindent 6 }}
+        {{- toYaml . | nindent 8 }}
         {{- end }}
       securityContext:
         {{- with .Values.securityContext }}


wrenix:
@danishVaid could you verify, that this patch solves that problem?
(with fluxcd it should be easy to take an branch/commitid from my fork-repo)

charts/nextcloud/templates/metrics-servicemonitor.yaml
```diff
@@ -30,10 +30,10 @@ spec:
...
-      interval: {{ .Values.metrics.serviceMonitor.interval }}
+      {{- with .Values.metrics.serviceMonitor.interval }}
+      interval: {{ . }}
       {{- end }}
-      {{- if .Values.metrics.serviceMonitor.scrapeTimeout }}
-      scrapeTimeout: {{ .Values.metrics.serviceMonitor.scrapeTimeout }}
+      {{- with .Values.metrics.serviceMonitor.scrapeTimeout }}
+      scrapeTimeout: {{ . }}
       {{- end }}
 {{- end }}
```
jessebot, line 39: Lets please try to keep PRs to only what's needed to fix this. If this is also broken, we can add it to the fix, but we should also detail that in the title and description of the PR.

charts/nextcloud/templates/metrics-service.yaml
```diff
@@ -9,18 +9,18 @@ metadata:
...
   type: {{ .Values.metrics.service.type }}
-  {{ if eq .Values.metrics.service.type "LoadBalancer" -}} {{ if .Values.metrics.service.loadBalancerIP }}
+  {{- if and (eq .Values.metrics.service.type "LoadBalancer") .Values.metrics.service.loadBalancerIP }}
   loadBalancerIP: {{ .Values.metrics.service.loadBalancerIP }}
-  {{ end -}}
-  {{- end -}}
+  {{- end }}
   ports:
   - name: metrics
     port: 9205
```
jessebot, line 26: same with this, please detail what you are fixing in the PR description or submit a seperate PR for this.

jessebot (changes requested):
I'll approve the workflow run after the comments are addressed, but thanks again for quickly taking note of the bug :)

Also, before I forget, you can test it by adding a custom config and then doing a `helm template .` from the helm chart directory of your nextcloud fork's patch branch. If you want to specify custom values, you can use `--values myvalues.file.yaml` and if you want to specify a directory to write the rendered templates out to for checking, you can use `--output-dir mytestdir`.

wrenix:
you are correct, after a second look i start to cleanup more - so i split it again.

i know for `helm template` command, but i do not deploy it - so the bug is more a php problemen then an helm.
We should  setup [chart-testing](https://github.com/helm/chart-testing-action), that try a deployen with kind and maintaine a example value.yaml file which enable all features in this helmchart

ishioni:
@wrenix https://github.com/ishioni/homelab-ops/blob/master/.github/workflows/flux-diff.yaml

DanishVaid:
Hmm, I'm trying to test but seeing an odd error during chart rendering (shown below). I don't think this is tied to your change. Seeing if I can figure out what it is

`Error: template: test/charts/postgresql/templates/primary/svc.yaml:11:14: executing "test/charts/postgresql/templates/primary/svc.yaml" at <include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $)>: error calling include: template: test/charts/mariadb/charts/common/templates/_labels.tpl:11:27: executing "common.labels.standard" at <include "common.names.name" .>: error calling include: template: test/charts/mariadb/charts/common/templates/_names.tpl:11:18: executing "common.names.name" at <.Chart.Name>: nil pointer evaluating interface {}.Name helm.go:84: [debug] template: test/charts/postgresql/templates/primary/svc.yaml:11:14: executing "test/charts/postgresql/templates/primary/svc.yaml" at <include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $)>: error calling include: template: test/charts/mariadb/charts/common/templates/_labels.tpl:11:27: executing "common.labels.standard" at <include "common.names.name" .>: error calling include: template: test/charts/mariadb/charts/common/templates/_names.tpl:11:18: executing "common.names.name" at <.Chart.Name>: nil pointer evaluating interface {}.Name`

wrenix:
@DanishVaid i am really interestest in your value.yaml / environment.

it look like it has nothing todo with this or #465 - that looks like an subchart problem (where you enable both, mariadb and postgresql at the same time? that does not look like the problem here, it is an recusive template part down to the sub-sub-chart and there the access to .Chart.Name)

DanishVaid:
Ah, the issue was definitely on my end. I spun up an extra deployment to ensure my main one wouldn't go down (being used by friends and family) and forgot to copy the base values so only my overlay was being applied. Can confirm that this PR fixes the issue I was seeing with the PHP config (new rendered configmaps shown below):

Screenshot 2023-11-19 at 2 10 32 PM

wrenix:
No worry, i take another look and find another bug (see podLabel commit), so thanks to get me spending timd to reviewe my own changes again xD

jessebot (approved):

PR #483, opened 2023-11-24
Include nextcloud service in trusted domains for metrics exporter

darkowlzz:
# Pull Request

## Description of the change

When nextcloud-exporter is used by enabling metrics, it fails to fetch the serverinfo with 400 response code.

nextcloud-exporter log
```console
level=error msg="Error during scrape: unexpected status code: 400"
```

nextcloud log
```console
10.244.0.38 - admin [24/Nov/2023:22:34:50 +0000] "GET /ocs/v2.php/apps/serverinfo/api/v1/info?format=json HTTP/1.1" 400 11344 "-" "nextcloud-exporter/0.6.1"
```

This is related to the switch from ingress to local service endpoint for nextcloud-exporter in https://github.com/nextcloud/helm/pull/388.

To fix this, in addition to `.Values.nextcloud.host`, the nextcloud service also needs to be included in the trusted domains.
In config/config.php
```php
  'trusted_domains' => 
  array (
    0 => 'localhost',
    1 => 'nextcloud.kube.home',
    2 => 'nextcloud',
  ),
```
where `nextcloud` is the service name.

This solution is based on https://github.com/xperimental/nextcloud-exporter/issues/14 , an upstream issue for the same problem.

## Benefits

The nextcloud-exporter will be able to fetch and export the metrics properly.

## Possible drawbacks

None

## Applicable issues

None

## Additional information

Without this, the metrics exporter exported metrics about itself only:

```
# HELP nextcloud_exporter_info Information about the nextcloud-exporter.
# TYPE nextcloud_exporter_info gauge
nextcloud_exporter_info{commit="98d4d4607e9d020dc19b8c21e1401b6b27a5d3a8",version="0.6.1"} 1
# HELP nextcloud_scrape_errors_total Counts the number of scrape errors by this collector.
# TYPE nextcloud_scrape_errors_total counter
nextcloud_scrape_errors_total{cause="other"} 2
# HELP nextcloud_up Indicates if the metrics could be scraped by the exporter.
# TYPE nextcloud_up gauge
nextcloud_up 0
```

Once the issue is fixed, it exports all the nextcloud app metrics:

```
# HELP nextcloud_active_users_daily_total Number of active users in the last 24 hours.
# TYPE nextcloud_active_users_daily_total gauge
nextcloud_active_users_daily_total 0
# HELP nextcloud_active_users_hourly_total Number of active users in the last hour.
# TYPE nextcloud_active_users_hourly_total gauge
nextcloud_active_users_hourly_total 0
# HELP nextcloud_active_users_total Number of active users for the last five minutes.
# TYPE nextcloud_active_users_total gauge
nextcloud_active_users_total 0
# HELP nextcloud_apps_installed_total Number of currently installed apps
# TYPE nextcloud_apps_installed_total gauge
nextcloud_apps_installed_total 42
# HELP nextcloud_apps_updates_available_total Number of apps that have available updates
# TYPE nextcloud_apps_updates_available_total gauge
nextcloud_apps_updates_available_total 0
...
```

## Checklist 


wrenix:
you need to set the complete domain, since that was merged:
https://github.com/nextcloud/helm/pull/505

(or we should use a public domain here).

so accept my suggestion it could be merged:

charts/nextcloud/templates/_helpers.tpl
```diff
@@ -178,7 +178,7 @@ Create environment variables used to configure the nextcloud container as well a
       name: {{ .Values.nextcloud.existingSecret.secretName | default (include "nextcloud.fullname" .) }}
       key: {{ .Values.nextcloud.existingSecret.passwordKey }}
 - name: NEXTCLOUD_TRUSTED_DOMAINS
-  value: {{ .Values.nextcloud.host }}
+  value: {{ .Values.nextcloud.host }} {{ if .Values.metrics.enabled }}{{ template "nextcloud.fullname" . }}{{ end }}
```
wrenix, line 181:
```suggestion
  value: {{ .Values.nextcloud.host }} {{ if .Values.metrics.enabled }}{{ template "nextcloud.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local{{ end }}
```

wrenix (changes requested):
you need to set the complete domain, since that was merged:
https://github.com/nextcloud/helm/pull/505

(or we should use a public domain here).

so accept my suggestion it could be merged:

charts/nextcloud/templates/_helpers.tpl
jessebot: does there need to be a comma in between the domains here? I'm not sure how this variable works, so I'm just clarifying. Then, I can sign off and commit the suggestion.

christensenjairus:
Here's a quick fix for those (like me) that can't wait for the chart to be fixed. This file's values will override those in config/config.php.

```yaml
nextcloud:
  configs: # I've taken the standard trusted domains and added one that's necessary for the exporter to work
    trusted-domains.config.php: |-
      <?php 
      $CONFIG = [
        'trusted_domains' =>
        array (
          0 => 'localhost',
          1 => 'nextcloud.example.com',
          2 => 'nextcloud.nextcloud.svc.cluster.local'
        ),
      ];
      ?>
```

charts/nextcloud/templates/_helpers.tpl
jessebot: Checked and no space needed:
https://github.com/nextcloud/docker?tab=readme-ov-file#auto-configuration-via-environment-variables

jessebot (approved):

PR #492, opened 2023-12-03
Update README.md - add section about two or more replicas with traefik

supermario18b:
# Pull Request

## Description of the change

It was explained how to create a traefik ingressRoute Resource that uses sticky sessions

## Benefits

Users can login when more nextcloud replicas are used

## Possible drawbacks

I don't see any

## Applicable issues

when more replicas


charts/nextcloud/README.md
```diff
@@ -30,6 +30,7 @@ helm install my-release nextcloud/nextcloud
     * [Downloading models for recognize](#downloading-models-for-recognize)
 * [Backups](#backups)
 * [Upgrades](#upgrades)
+* [Two or more replicas with traefik](https://github.com/nextcloud/helm/tree/main/charts/nextcloud#Two-or-more-replicas-with-traefik)
```
jessebot, line 33:
```suggestion
* [Two or more replicas with traefik](#Two-or-more-replicas-with-traefik)
```
This should do the trick with a smaller link :)

wrenix:
Maybe we should use annotations and write docu for multiple ingress-controller:
   - here for traefik: https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/#on-service
   - here for nginx: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#session-affinity
   - here for haproxy: https://haproxy-ingress.github.io/docs/configuration/keys/#affinity
   
 Your docu is very traefik specifiy and would be maybe replaced with gateway-api and standard over multiple ingress-controller:
 https://gateway-api.sigs.k8s.io/implementations/

if there exists something for session-sticky: https://github.com/kubernetes-sigs/gateway-api/issues/1619

supermario18b:
> Or @supermario18b do you like to improve there the traefik section?
you can close the PR

PR #496, opened 2023-12-18
feat: add support for immutable image

thefirstofthe300:
# Pull Request

## Description of the change

Nextcloud is capable of being built using an immutable image (see https://github.com/thefirstofthe300/nextcloud-docker). In this scenario, there is no point in mounting the `/var/www` and `/var/www/html` directories to store state. The remaining directories should be mounted, but not these two.

## Benefits

Allows users to build their own immutable images which do not perform an rsync on start but still use this chart.

## Possible drawbacks

The default values are not changing so I see no drawbacks.

## Additional information


provokateurin:
Nextcloud is not really immutable. It will probably run into quite a few issues because all code assumes the whole dir is writable.

mcronce:
@provokateurin sorry for necroposting, but I couldn't find an answer searching in here or the the docker/server repos - is there any effort being made to change that?  Running with the entire codebase in shared writable persistent storage makes for a pretty bad experience; those types of storage tend to be (necessarily, for consistency across networks) quite slow.  There's no reason that config files can't live in mutable storage, but I wouldn't think the PHP code would need to overwrite itself (or HTML/JS/CSS, for that matter)

PR #513, opened 2024-01-26
fix #489 resources indent in README

### LINKED ISSUE #489: Manual indentation is incorrect
https://github.com/nextcloud/helm/blob/8cb6585289f1fd9533233f44e6b2fb01dd85b82f/charts/nextcloud/README.md?plain=1#L426

The correct indentation value here is 0

Deaddy:
# Pull Request

## Description of the change

We remove an indent in the README.md so it matches the values file and the reference in the corresponding helm template file.

## Benefits

The README.md matches the code.

## Possible drawbacks

In general one could argue that having `resources:` outside of the `nextcloud:`-Block in the values is not consistent with other components of the deployment, but for the sake of compatibility fixing the README seems to be appropriate.

## Applicable issues

- fixes #489

## Additional information

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.11
+version: 4.5.12
 appVersion: 28.0.1
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


provokateurin (approved):
Thanks for the fix, I agree that the resources should be under the nextcloud key instead, but that can be fixed later.

PR #481, opened 2023-11-20
fix: cleanup and linting of metric resources

wrenix:
# Pull Request

## Additional information

Split of #478 

1. Move the metrics-deployment related resources into dedicated folder `templates/metrics`.
  make a fee cleanups and lints for easier readable:
     - indent of list
     - use nindent instatt of indent
     - use with instatt of if
     - merge if with using `and` (`and` could handle more then two parameters)

2. Add namespaceSelector on servicemonitor

3. add `ci/*-values.yaml` for [chart-testing](https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md)

 
## Checklist 

- [x] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).

PR #478, opened 2023-11-19
fix: cleanup metrics components (and move to folder)

wrenix:
# Pull Request

## Applicable issues

- fixes #

## Checklist 

- [ ] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] (optional) Variables are documented in the README.md

PR #516, opened 2024-02-02
chore: move metrics yaml to extra folder

wrenix:
a first / last try of a really easy reviewable PR

if you see this and need more then 5min ....

@provokateurin 

a split of #481 which is an split of #478  (which does not contain any changes in logic and is 2,5 month old)


provokateurin:
I understand you are frustrated, but please stay friendly towards others https://nextcloud.com/contribute/code-of-conduct/

provokateurin:
Honestly I'm not sure if this change makes sense. According to https://helm.sh/docs/chart_best_practices/templates/ the current structure is best practice and the page doesn't mention that subdirectories are allowed (but it also doesn't say they are not allowed). https://stackoverflow.com/a/77619725 suggests that it still works, even if it is not documented.

wrenix:
your first link is good, but it does not say anything about, if much of `my-example` exists and nothing about folders inside of templates.

your stackoverflow post, say - that it works (and yes it works also for application charts).

provokateurin (approved):
Either way, I see that is change makes sense to make the structure less cluttered.

wrenix:
that change does not change any logic -> semver a chore, which does not need a new version, but maybe your ci failed ...

provokateurin:
CI is still running, I'm not sure if it will require a new version :thinking:

PR #480, opened 2023-11-20
fix: improve handling of config files

wrenix:
Split of #478 

Move all the configfiles for nginx, nextcloud, htaccess out of the kubernetes-resources as single file beside the helm chart and load them with helm function into the kubernetes-resources.

benefits:
  - config files are readable (still support of golang/sring/helm-templates)
  - logic in which `defaultConfigs` is loaded is generic (it is a `range` over the map)

PR #517, opened 2024-02-02
easy readible indent

wrenix:
# Pull Request

array with:
```yaml
array:
  - 3
```
instatt of
```yaml
array:
- 3
```

+ 
this is just of templating part `{{-` will remove all spaces before that, so moved to level of yaml structure.

(i change of `{{` to `{{-` will happen in another PR with the replace of `if` to `when`).

draft, i am waiting for the merge of #516

### DIFF
--- a/charts/nextcloud/templates/ingress.yaml
+++ b/charts/nextcloud/templates/ingress.yaml
@@ -9,37 +9,37 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/component: app
-{{- if .Values.ingress.labels }}
+    {{- if .Values.ingress.labels }}
 {{ toYaml .Values.ingress.labels | indent 4 }}
-{{- end }}
-{{- if .Values.ingress.annotations }}
+    {{- end }}
+  {{- if .Values.ingress.annotations }}
   annotations:
 {{ toYaml .Values.ingress.annotations | indent 4 }}
-{{- end }}
+  {{- end }}
 spec:
   {{- if .Values.ingress.className }}
   ingressClassName: {{ .Values.ingress.className }}
   {{- end }}
   rules:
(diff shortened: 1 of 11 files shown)


charts/nextcloud/templates/ingress.yaml
```diff
@@ -9,37 +9,37 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/component: app
-{{- if .Values.ingress.labels }}
+    {{- if .Values.ingress.labels }}
 {{ toYaml .Values.ingress.labels | indent 4 }}
```
provokateurin, line 13:
```suggestion
    {{- toYaml .Values.ingress.labels | indent 4 }}
```
Wouldn't it make sense to also indent this for better readability?

provokateurin: Or am I misunderstanding you and you want to do this in a follow-up PR?

provokateurin:
Do you know if there is a tool that could automatically enforce these style rules? I think your suggested changes are very good, I just want to ensure that they will stay that way and that other places can be safely refactored as well.

charts/nextcloud/templates/ingress.yaml
wrenix: in the following PR (and there replace `if` with `with`)

wrenix:
At the moment, i do not know any tools - i will take a look into the chart-testing you already use (it use in background the yamllint tool written in python)

provokateurin (approved):

provokateurin:
Shall we merge this already or wait for a potential tool that could enforce the style?

wrenix:
i like to create the next PR ... in the end, i want also add some chart-testing values-files as best practics ... maybe we discuss it there.

e.g.: https://github.com/nextcloud/helm/pull/480/commits/a6ca7bd23c7bcaa59816edee69b1156a8fc001ec

PR #518, opened 2024-02-02
chore: replace if with with (and improve strip using with -)

wrenix:
# Pull Request

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.5.17
+version: 4.5.18
 appVersion: 28.0.2
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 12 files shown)


provokateurin (commented):
LGTM, but I think in a few places nindent is used where indent would be enough. Could you fix that?

wrenix:
I am a hours away from keyboard, sry -> i believe there is needed, but i take a look

wrenix:
you are correct, if i would not use in some places `{{- toYaml`, i could use `indent` instatt of `nindent`.
I often use it together with `{{- with` and {{- end`. so it looks for me nice on writing and reading.

that is a kind of personal preference. should i change it?

provokateurin:
I  see, that makes sense. Then I prefer this style :) Still, it would be great to have a tool to enforce the style.

wrenix:
increase chart-version after rebase

provokateurin (approved):

Issue #466, opened 2023-11-11
Manage apps with initContainer


wrenix:
i like to add a easy mechanis to setup and configure appplication.

so i see the following new options in `values.yaml`:

```yaml 
nextcloud:
  appManagement:
    # -- if app not listed under apps it will be disabled
    autoDisable: true
    # -- if app not listed under apps it will be uninstalled
    autoUninstall: true
    apps:
       deck:
         enabled: true
       drawio:
          enabled: true
          config:
            DrawioUrl: "https://embed.diagrams.net"
            DrawioOffline: "yes"
```

Therefore
- i will use the occ command with an initContainer for enable and check there installation
- `nextcloud.appManagement.apps` is an dict for easy use multiple `value.yaml` files and merge them
 - config of applications are stored in an kubernetes-secret because they could containe sensitiv informations

jessebot:
This does sound pretty neat and streamlined :eyes: I think it's a good idea. How about you friends, @provokateurin @tvories ? :)

provokateurin:
I like the idea as well, I wonder how much work it is to concert the yaml into the php array syntax with nesting etc. (and how easy it could break).

jessebot:
> I like the idea as well, I wonder how much work it is to concert the yaml into the php array syntax with nesting etc. (and how easy it could break).

yeah, that makes sense. I think my main concern is the config section with each app. It may make more sense to start with support for just installing, enabling, and disabling, as config can get hairy, as you mentioned with regards to how nested a config for a given app might be, especially considering some apps will have different methods for config than others :thinking:

wrenix:
That is an implementation detail, that i like to discuss in my PR later.
but if you like to know - i wanted to write shell script and with using occ config:set (or config:import) without writing any php code ( and handling arrays in phpo.

provokateurin:
Sounds good, but how would you delete keys that got removed from the config?

jessebot:
When this PR, https://github.com/nextcloud/docker/pull/2115, get merged, we could use the docker level hooks by mounting configMaps into the `post-installation` or `before-starting` folders as per the [docs](https://github.com/nextcloud/docker#auto-configuration-via-hook-folders). It was discussed a bit in https://github.com/nextcloud/docker/issues/763#issuecomment-1857453976 as to why we can't do that for now, but knowing these folders exist, reduces the need for more init containers or external post install jobs, not just for managing apps, for running other `occ` commands as well.

provokateurin:
[...]

PR #525, opened 2024-02-04
feat: add hook scripts to pod

wrenix:
# Pull Request

## Description of the change

first step or prepare to #466 

thanks for this hint @jessebot 
https://github.com/nextcloud/helm/issues/466#issuecomment-1857616753

## Benefits

not needed to create a own secret and complex mount per extraVolumes needed to setup such hooks like in:
https://github.com/nextcloud/docker#auto-configuration-via-hook-folders

descripted 

## Checklist 

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -147,6 +147,16 @@ nextcloud:
   #        )
   #      );
 
+  # Hooks for auto configuration
+  # Here you could write small scripts which are placed in `/docker-entrypoint-hooks.d/<hook-name>/helm.sh`
+  # ref: https://github.com/nextcloud/docker?tab=readme-ov-file#auto-configuration-via-hook-folders
+  hooks:
+    pre-installation:
+    post-installation:
+    pre-upgrade:
+    post-upgrade:
+    before-starting:
+
   ## Strategy used to replace old pods
   ## IMPORTANT: use with care, it is suggested to leave as that for upgrade purposes
   ## ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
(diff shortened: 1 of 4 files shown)


charts/nextcloud/values.yaml
```diff
@@ -146,7 +146,15 @@ nextcloud:
   #          )
   #        )
   #      );
-
+  # Hooks for auto configuration
+  # Here you could write small scripts which are placed in `/docker-entrypoint-hooks.d/<hook-name>/<hook-name>.sh`
```
provokateurin, line 150:
```suggestion
  # Here you could write small scripts which are placed in `/docker-entrypoint-hooks.d/<hook-name>/helm.sh`
```
They are always mounted like this, no?

charts/nextcloud/templates/secrets-hooks.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if not (values .Values.nextcloud.hooks | compact | empty) }}
+---
+apiVersion: v1
+kind: Secret
```
provokateurin, line 4: Can you explain why you chose Secret over a simple ConfigMap?

charts/nextcloud/values.yaml
wrenix: Correct (i change it after writing docu)

charts/nextcloud/templates/secrets-hooks.yaml
wrenix: I like to write an script/hook to setup oidc (which contains client-secret).

or other tokens and secrets could be in the scripts

wrenix: (but yes i could change it to ConfigMap and use therefore environment variables)

provokateurin: I see. I think this should just be a ConfigMap and then you can use the environment variables for the real secrets.

provokateurin (approved):

provokateurin:
Just need to solve the conflicts

PR #529, opened 2024-02-05
fix: another cleanup and linting

wrenix:
# Pull Request

## Description of the change
Left over in my branches

## Benefits

nothing

## Possible drawbacks

nothing

## Additional information

not importend - does not have to be merged
... i could split it also in some more PRs.

But i will focus before on #480

## Checklist 


charts/nextcloud/values.yaml
```diff
@@ -3,7 +3,9 @@
 ##
 image:
   repository: nextcloud
-  # tag: 24.0.3-apache
+  flavor: apache
```
provokateurin, line 6: I remember in the past there was some back and forth about where the default flavor should be defined. I think having it in the values.yaml file makes the most sense.

charts/nextcloud/values.yaml
```diff
@@ -3,7 +3,9 @@
...
+  # -- tag: 24.0.3-apache
```
provokateurin, line 7:
```suggestion
  # tag: 24.0.3-apache
```
Or do these serve any purpose?

provokateurin (changes requested):

provokateurin (approved):

charts/nextcloud/values.yaml
jessebot: If we used helm-docs, that would be cool. It looks like they recently added the ability to do sections!
https://github.com/norwoodj/helm-docs/tree/master/example-charts/sections

I'll open a seperate issue for this.

PR #530, opened 2024-02-05
feat: appManagement per helm values

### LINKED ISSUE #466: Manage apps with initContainer

wrenix:
# Pull Request

## Benefits

Easy Helm-Values management for nextcloud apps

## Applicable issues

- fixes #466 

## Checklist 

### DIFF
--- a/charts/nextcloud/files/app-management.sh.gotmpl
+++ b/charts/nextcloud/files/app-management.sh.gotmpl
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+{{- range $app,$appConfig := .Values.nextcloud.appManagement.apps }}
+echo "== manage app: {{ $app }} ==";
+
+{{- if $appConfig.enabled }}
+echo "enable:"
+/var/www/html/occ app:install {{ $app }}
+/var/www/html/occ app:enable {{ $app }}
+
+# config of app
+echo "config:"
+{{- range $key, $value := $appConfig.config }}
+/var/www/html/occ config:app:set -n {{ $app }} {{ $key }} --value="{{ $value }}"
+{{- end}}
+
+{{- else }}
+echo "disable:"
+/var/www/html/occ app:disable {{ $app }}
+{{- end }}
+{{- end }}{{/* end-range appManagement.apps */}}
+
+# TODO autoDisable and autoRemove use occ app:list
+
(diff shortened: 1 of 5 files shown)


charts/nextcloud/files/app-management.sh.gotmpl
```diff
@@ -0,0 +1,24 @@
...
+
+{{- if $appConfig.enabled }}
+echo "enable:"
+/var/www/html/occ app:install {{ $app }}
+/var/www/html/occ app:enable {{ $app }}
+
+# config of app
+echo "config:"
+{{- range $key, $value := $appConfig.config }}
+/var/www/html/occ config:app:set -n {{ $app }} {{ $key }} --value="{{ $value }}"
```
provokateurin, line 14: In some cases you want to be able to specify the value type, otherwise `"true"` will be set as a string and not as a boolean.

provokateurin (commented):
I'm still not sure if this is a good idea.

wrenix:
I am also not ... maybe i put that into my hook script like in #525 ...

I just to show the result here.


PS: Thank you for the review marathon the last days. I am really sorry about to being so rough / frustrated.

charts/nextcloud/values.yaml
```diff
@@ -157,6 +157,21 @@ nextcloud:
...
 
+  appManagement:
+    enabled: false
+    # -- if app not listed under apps it will be disabled
+    autoDisable: true
+    # -- if app not listed under apps it will be uninstalled
+    autoRemove: true
+    apps:
+       deck:
+         enabled: false
```
jessebot, line 168:
```suggestion
    apps:
      deck:
        enabled: false
```

I think this fixes the indentation issue? 🤔

PR #534, opened 2024-02-10
Add nodeSelector, tolerations and affinity to metrics deployment

SwitzerChees:
Signed-off-by: Patrick Michel <patrickmichel96@yahoo.de>


SwitzerChees:
Just added the options nodeSelector, tolerations and affinity to the metrics template. For our use case it is important to also be able to control the placement of this pod based on these rules. And probably other people will also be happy about that in the future.

provokateurin (commented):
LGTM, maybe @wrenix you also want to take a look? :)

provokateurin:
You need to bump the chart version btw.

SwitzerChees:
> You need to bump the chart version btw.

Done just a patch version.
Or is it a minor version do you follow the semantic versioning concept? ;-)

charts/nextcloud/values.yaml
```diff
@@ -508,6 +508,12 @@ metrics:
 
   # podLabels: {}
 
+  # nodeSelector: {}
```
wrenix, line 511: Want do we like, to use that without comment out?
```suggestion
  nodeSelector: {}
```

wrenix (approved):
The code part looks really nice.

For my opionion, it is such a small new feature, that i would use a patch level.
But yes, you are right that is a feature so in semver it would be a minor.

charts/nextcloud/values.yaml
provokateurin: I have no strong opinion, but in case we do this we could also drop the `with` part because we always know that there will be a value. Maybe a general consideration we should think about and then implement in the whole chart.

SwitzerChees: Just let me know when you decided how to handle this in general and then I can update my pull requests so that it fits your requirements.

provokateurin: Ah ok, I didn't know that. Then I prefer not commenting it out like you suggested.

SwitzerChees: Ok I commented in all three properties but if this is suggested there are also a lot of other properties like podLabels or podAnnotations and they are still commented out? So not very consistent at the moment.

jessebot: I wanted briefly chime in and to give my vote for having these options uncommented and then using `with` whereever they're templated, that way the values.yaml is more clear. Yeah, it needs to be fixed in a lot of places, but worrying about just the stuff for this PR makes the most sense in the moment. Other PRs can tackle the other places it needs fixing later.

charts/nextcloud/values.yaml
```diff
@@ -62,10 +62,10 @@ ingress:
   path: /
   pathType: Prefix
 
-
 # Allow configuration of lifecycle hooks
 # ref: https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
-lifecycle: {}
+lifecycle:
+  {}
```
jessebot, line 68:
```suggestion
lifecycle: {}
```

charts/nextcloud/values.yaml
```diff
@@ -426,7 +426,8 @@ persistence:
     accessMode: ReadWriteOnce
     size: 8Gi
 
-resources: {}
+resources:
+  {}
```
jessebot, line 430:
```suggestion
resources: {}
```

provokateurin:
Something looks off with all these commits, please rebase onto master to drop all of them.

PR #538, opened 2024-02-22
Fix subPath on datadir mounts (fixes #531)

### LINKED ISSUE #531: Unable to mount datadir without `/data` suffix
## Describe your Issue

At the moment it is impossible to mount a data directory which is not $somepath/data.

The offending code is in [lines 218-226 of deployment.yaml](https://github.com/nextcloud/helm/blob/main/charts/nextcloud/templates/deployment.yaml#L218-L226):

```
            {{- if and .Values.persistence.nextcloudData.enabled .Values.persistence.enabled }}
            - name: nextcloud-data
              mountPath: {{ .Values.nextcloud.datadir }}
              subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.nextcloudData.subPath) (empty .Values.persistence.nextcloudData.subPath) }}
            {{- else }}
            - name: nextcloud-main
              mountPath: {{ .Values.nextcloud.datadir }}
              subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.subPath) (empty .Values.persistence.subPath) }}
            {{- end }}
```

We see that `/data` is being appended irrespective of the content or existence of subPath.

Another issue is that the data dir is only used if also `.Values.persistence.enabled` is enabled and it would be convenient to remove this unnecessary `and`-condition in the same PR, but I am not yet familiar with how atomic we like our PRs here. :-)

## Approaches

I think there are basically two ways to deal with this.

1. The cleaner version would be to set `/data` as the subPath-default Value, so it would only break for people who have set a subPath and they would need to update their values accordingly.

2. The not so nice but non-breaking change would be to add another boolean variable with a glorious name like `postFixData` which is by default set to true and would then use the old logic, and if it is not set to true it would drop the subPath if empty and not add `/data` if it is set.

So, which approach do we prefer or did I possibly even miss a trick to make this work without code change?

Deaddy:
- subPath now does not necessarily end on /data, in particular empty subPath possible for datadir on specific PVC
- datadir persistence now does not depend on overall persistence anymore
- bump .minor because old installs will need to add `/data` back to the subPath in the values file

# Pull Request

## Description of the change

Fix #531

The datadir had the subPath-Suffix "/data" forced upon it and it was impossible 

## Possible drawbacks

If someone actually used the a different volume in the past, they will now need to add `/data` to their subPath value.

## Applicable issues

- fixes #531 

## Checklist 

- [x] tested installing and running fresh nextcloud instance with existing persistent volume for datadir

### DIFF
--- a/charts/nextcloud/templates/_helpers.tpl
+++ b/charts/nextcloud/templates/_helpers.tpl
@@ -246,11 +246,13 @@ Create volume mounts for the nextcloud container as well as the cron sidecar con
 - name: nextcloud-main
   mountPath: /var/www/html
   subPath: {{ ternary "html" (printf "%s/html" .Values.nextcloud.persistence.subPath) (empty .Values.nextcloud.persistence.subPath) }}
-{{- if and .Values.persistence.nextcloudData.enabled .Values.persistence.enabled }}
+{{- if .Values.persistence.nextcloudData.enabled }}
 - name: nextcloud-data
   mountPath: {{ .Values.nextcloud.datadir }}
-  subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.nextcloudData.subPath) (empty .Values.persistence.nextcloudData.subPath) }}
-{{- else }}
+{{- if not (empty .Values.persistence.nextcloudData.subPath) }}
+  subPath: {{ .Values.persistence.nextcloudData.subPath }}
+{{- end }}
+{{- else if .Values.persistence.enabled }}
 - name: nextcloud-main
   mountPath: {{ .Values.nextcloud.datadir }}
   subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.subPath) (empty .Values.persistence.subPath) }}
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.7.0
+version: 5.0.0
 appVersion: 28.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.2
+version: 4.7.0
 appVersion: 28.0.2
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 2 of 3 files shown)


charts/nextcloud/templates/_helpers.tpl
```diff
@@ -246,11 +246,13 @@ Create volume mounts for the nextcloud container as well as the cron sidecar con
...
   subPath: {{ ternary "html" (printf "%s/html" .Values.nextcloud.persistence.subPath) (empty .Values.nextcloud.persistence.subPath) }}
-{{- if and .Values.persistence.nextcloudData.enabled .Values.persistence.enabled }}
+{{- if .Values.persistence.nextcloudData.enabled }}
 - name: nextcloud-data
   mountPath: {{ .Values.nextcloud.datadir }}
-  subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.nextcloudData.subPath) (empty .Values.persistence.nextcloudData.subPath) }}
-{{- else }}
+{{- if not (empty .Values.persistence.nextcloudData.subPath) }}
+  subPath: {{ .Values.persistence.nextcloudData.subPath }}
+{{- end }}
```
provokateurin, line 254: This is a breaking change that deserves a bump of the major version

provokateurin (commented):
I think an alternative that doesn't require a breaking change would be to add another value that specifies the path of the data folder inside the subpath and if that is empty it is just omitted.

charts/nextcloud/templates/_helpers.tpl
Deaddy: Oh I mixed up minor and major, fixed.

I do not really like adding values but am not defensive about it, if you would prefer that.

wrenix (changes requested):

joaofasilva-kitar:
bump

charts/nextcloud/templates/_helpers.tpl
```diff
@@ -246,11 +246,13 @@ Create volume mounts for the nextcloud container as well as the cron sidecar con
 - name: nextcloud-main
   mountPath: /var/www/html
   subPath: {{ ternary "html" (printf "%s/html" .Values.nextcloud.persistence.subPath) (empty .Values.nextcloud.persistence.subPath) }}
-{{- if and .Values.persistence.nextcloudData.enabled .Values.persistence.enabled }}
+{{- if .Values.persistence.nextcloudData.enabled }}
 - name: nextcloud-data
   mountPath: {{ .Values.nextcloud.datadir }}
-  subPath: {{ ternary "data" (printf "%s/data" .Values.persistence.nextcloudData.subPath) (empty .Values.persistence.nextcloudData.subPath) }}
-{{- else }}
+{{- if not (empty .Values.persistence.nextcloudData.subPath) }}
```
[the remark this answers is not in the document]
wrenix: ping

PR #540, opened 2024-02-28
Added support for Built-in CODE Server (ARM64)

dgiebert:
# Pull Request

## Description of the change

Made the nginx regex also catch the `_arm64` for the code server

## Additional information

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.2
+version: 4.6.3
 appVersion: 28.0.2
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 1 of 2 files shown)


provokateurin:
Thanks for your contribution, can you link to the code that shows this is necessary? It looks like https://github.com/nextcloud/documentation/blob/e9c1af1ff00625cb24e7e6ccf4d547c15bd714c8/admin_manual/installation/nginx-root.conf.sample#L145 also doesn't have it, could you also contribute it to the general documentation?

dgiebert:
It's a minor thing. Do you want me to bump the chart version?

provokateurin:
Yes our CI requires every PR that touches the templates to update the version. You can just bump the patch version though since it is only a bug fix.
Can you still link to the code that shows exactly why this change is needed?

dgiebert:
I will bump the version tomorrow then.

Sadly I cannot find it in code right now, but there are multiple pointers (and ofc I have tested it locally)

See here:  https://github.com/CollaboraOnline/richdocumentscode/issues/204#issuecomment-1517211152

Might be related to the app being called different, also see here: https://github.com/CollaboraOnline/richdocumentscode?tab=readme-ov-file#implementation

provokateurin:
I see, thanks a lot. Only bumping the version is needed, then we can merge this.

provokateurin (approved):

PR #542, opened 2024-03-03
Add backup cronjobs

stephan2012:
# Pull Request

## Description of the change

Allow creation of Kubernetes Cronjobs for periodic backups of the Nextcloud data volume.

## Benefits

The PR provides a standard method of installing backup cronjobs. It uses the Helm chart's configuration data to share the configuration.

## Possible drawbacks

I have yet to be aware of any drawbacks.

## Applicable issues

N/A

## Checklist 


provokateurin:
Cool idea, I hope to be able to review it soon.
I'm wondering if it makes sense to have the extra values both in the readme and the values.yaml as they tend to get out of sync at some point.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.3
+version: 4.7.0-1
```
```suggestion
version: 4.7.0
```
I think calling it 4.7.0 is fine. You may also need to rebase, as the appVersion is out of date here.

charts/nextcloud/README.md
```diff
@@ -11,26 +11,34 @@ helm install my-release nextcloud/nextcloud
...
-* [Hugepages](#hugepages)
-* [HPA (Clustering)](#hpa-clustering)
-* [Running `occ` commands](#running-occ-commands)
-    * [Putting Nextcloud into maintanence mode](#putting-nextcloud-into-maintanence-mode)
-    * [Downloading models for recognize](#downloading-models-for-recognize)
-* [Backups](#backups)
-* [Upgrades](#upgrades)
-* [Troubleshooting](#troubleshooting)
+- [TL;DR;](#tldr)
+- [Quick Links](#quick-links)
```
jessebot, line 15:
```suggestion
```
Quick links is already the header we're in, so I think this can be removed

stephan2012:
Oops. Forgot about this.

I’ll do the rebase today. Not sure about maintenance mode … I’d prefer to leave the system online and do a file scan after restore.

stephan2012:
> maintenance mode is recommended to be on during backups: https://docs.nextcloud.com/server/latest/admin_manual/maintenance/backup.html

@jessebot For me, the only feasible way seems to be to mount the `config` folder into the backup container to allow modifications to `config/config.php`. Some logic in the backup container would be responsible for updating the file before and after the backup. For this option, the only required Helm chart extension would be the additional volume mount.

While Kubernetes has init containers (enter backup mode), it does not have the concept of "exit containers" (leave backup mode). Also, lifecycle hooks are not an option because [Kubernetes does not guarantee their execution time](https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion).

Thoughts?

charts/nextcloud/Chart.yaml
stephan2012: Updated to 5.5.0-1 for now.

charts/nextcloud/README.md
stephan2012: This is auto-generated; manually changing it would be cumbersome.

stephan2012:
> Actually, the way I do this is via just having [Kubernetes CronJobs for just the maintenance mode](https://github.com/small-hack/argocd-apps/tree/main/nextcloud/maintenance_mode_cronjobs) being turned on and off, […]

There are probably more reliable approaches than relying on wall-clock-based synchronization. It either reduces service availability or toggles the switch too early.

> and then I use a [k8up](https://k8up.io/) [`Schedule`](https://docs.k8up.io/k8up/2.8/references/schedule-specification.html), which is a scheduled restic backup of all the annotated PVCs in the nextcloud namespace, to actually do the backup. During that time, I also have a [scheduled backup](https://cloudnative-pg.io/documentation/1.23/backup/#scheduled-backups) that takes place via the [Cloud Native PostgreSQL Operator](https://cloudnative-pg.io/documentation/1.23/).

My preferred option is to use logging and continuous WAL archiving as provided by the [Zalando Postgres Operator](https://github.com/zalando/postgres-operator) (uses [wal-e](https://github.com/wal-e/wal-e) or [wal-g](https://github.com/wal-g/wal-g) under the covers), allowing you to restore the database consistently into any point in time.

> In my personal opinion, I think using a backup operator that wraps restic (Like [Velero](https://velero.io/) or K8up) is ideal, and the main thing we should be concerned about here is the maintenance mode cronjobs, but I'm also interested in hearing more opinions from other collaborators and maintainers such as @tvories and @provokateurin 🙏

Let's be agnostic about the backup tool. People will probably use a wide variety of backup solutions and technologies.

PR #549, opened 2024-04-01
Add: Redis persistent volumes settings to values

### LINKED ISSUE #546: Feature: Add persistence settings for redis
## Description of the change

Add persistence settings for the redis database. Such as storageClass and wether it should be enabled at all.

## Benefits

- It would be possible to use a different storageClass than default which would enable:
  - different type of storage
  - different backup settings
  - etcetera
- It would be possible to disable persistent storage. Since redis is only used as cache. One might opt to not use persistent storage at all.

## Possible drawbacks

- I can't think of any.

f3k-freek:
# Pull Request

## Description of the change

Added Redis persistent volume settings to values and the readme. This is only to make users aware of how to change these settings, the PR itself doesn't contain any changes to the functioning of the chart.

## Benefits

It will make people aware of how to change the persistent volume settings for Redis.

## Possible drawbacks

Too many settings in values might "clutter" the values file and readme.

## Applicable issues

- fixes #546

## Additional information

not applicable

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.7.0
+version: 4.6.6
 appVersion: 28.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.5
+version: 4.7.0
 appVersion: 28.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -355,6 +355,15 @@ redis:
     existingSecret: ""
     # Password key to be retrieved from existing secret
     existingSecretPasswordKey: ""
+  # Since Redis is used for caching only, you might want to use a storageClass with different reclaim policy and backup settings
+  global:
+    storageClass: ""
+  master:
+    persistence:
+      enabled: true
+  replica:
+    persistence:
+      enabled: true
 
 
 ## Cronjob to execute Nextcloud background tasks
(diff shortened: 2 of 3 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.5
+version: 4.7.0
```
provokateurin, line 3: IMO this is only a patch because it adds docs but no new features or fixes.

f3k-tech: Thank you for your fast review. I agree and changed it.

charts/nextcloud/values.yaml
```diff
@@ -355,6 +355,15 @@ redis:
...
     existingSecretPasswordKey: ""
+  # Since Redis is used for caching only, you might want to use a storageClass with different reclaim policy and backup settings
+  global:
+    storageClass: ""
+  master:
+    persistence:
+      enabled: true
+  replica:
+    persistence:
+      enabled: true
```
jessebot, line 366: The default should be for persistence to be off, and then users can enable it if they'd like, that way quick testing is less of a hassle to clean up.

charts/nextcloud/values.yaml
```diff
@@ -355,6 +355,15 @@ redis:
     existingSecret: ""
     # Password key to be retrieved from existing secret
     existingSecretPasswordKey: ""
+  # Since Redis is used for caching only, you might want to use a storageClass with different reclaim policy and backup settings
+  global:
+    storageClass: ""
+  master:
+    persistence:
+      enabled: true
```
jessebot, line 363:
```suggestion
      enabled: false
```

charts/nextcloud/values.yaml
jessebot, line 366:
```suggestion
      enabled: false
```

f3k-tech: I agree that persistence should be off by default. But in the current chart it's enabled by default. So if we disable it now, it might lead to unpleasant surprises for people who are relying on the default values.

So given that it's currently enabled by default, I would suggest to keep it that way.

jessebot (approved):

PR #551, opened 2024-04-02
Feature/client push

### LINKED ISSUE #109: High Performance File Backend
Are you planning to include the High Performance File Backend aka Client Push in the Helm chart or would you accept a pull request that adds this? This might be related to #106 as Nextcloud 21 is a base requirement for that.

AndreKoepke:
# Pull Request

## Description of the change

Add a container for client_push app. It allows websocket-connections for updates instead of polling. See [notify_push](https://github.com/nextcloud/notify_push)

## Benefits

* notify_push is usable

## Possible drawbacks

* Nextcloud-Chart must be started without notify_push, so it can be installed.
* Then Next-Chart can be started with notify_push
* After that, notify_push must be manually activated by running `sudo -u www-data ./occ notify_push:setup https://NEXTCLOUD_HOST/push`

## Applicable issues

- fixes #109

## Checklist 


AndreKoepke:
Can we install apps with commands? If so, it can be added as default.

charts/nextcloud/templates/deployment.yaml
```diff
@@ -274,6 +274,25 @@ spec:
           volumeMounts:
             {{- include "nextcloud.volumeMounts" . | trim | nindent 12 }}
         {{- end }}{{/* end-if cronjob.enabled */}}
+        {{ if .Values.clientPush.enabled }}
+        - name: {{ .Chart.Name }}-client-push
+          image: {{ include "nextcloud.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+            - custom_apps/notify_push/bin/x86_64/notify_push
```
AndreKoepke, line 282: Hard-coded `x86_64`

Leptopoda:
> Can we install apps with commands? If so, it can be added as default.

there is `occ app:install notify_push`
https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/occ_command.html#apps-commands-label

wrenix:
I really like the idea to run clientpush.
But please in a different way:
  - extra deployment (like for metrics)
  - extra service and ingress with Additional path

The reason: 
 - On this way, we could scale nextcloud indepenten fron push client
 -  And do not maintain nginx or Apache code, that could handle ingress-controller.
 

Another reason (i need to verify / hope that it exists), we could use a other image for the client push (without any distro inside thanks Rust) and trage a more beautiful/helm-like way the version

AndreKoepke:
> i start to write, what i mean under: #581 if you like to test it or use that as first state to improve your PR take it
I like your changes. I'm running my variant since I opened the PR, but I will switch to your variant in the next days.

Should we (or I) close my PR?

wrenix:
Till there is a solution merged i whould keep it open

charts/nextcloud/templates/deployment.yaml
```diff
@@ -85,10 +85,10 @@ spec:
           livenessProbe:
             httpGet:
               path: /status.php
-              port:  {{ $.Values.nextcloud.containerPort }}
+              port: {{ $.Values.nextcloud.containerPort }}
```
jessebot, line 88: Is this `$` supposed to be there?

AndreKoepke: I didn't inserted it. But yes, the `$` is not necessary there.

jessebot: oh, yeah, you're right. It's because it's within a `with`. Sorry, my bad 🤦

PR #554, opened 2024-04-09
chore: Prevent displaying constant diffs for nextcloud-token in ArgoCD

kahirokunn:
# Pull Request

## Description of the change

When metrics are not enabled, the use of randAlphaNum was causing constant diffs for nextcloud-token in ArgoCD. To mitigate this issue, now nextcloud-token won't be configured when metrics are not enabled.

## Benefits

This change prevents unnecessary diffs in ArgoCD when metrics are not enabled, leading to a cleaner deployment process.

## Possible drawbacks

None

## Applicable issues

- fixes #<issue_number>

## Additional information

None

## Checklist

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.5
+version: 4.7.0
 appVersion: 28.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.7.0
+version: 4.6.8
 appVersion: 29.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/templates/secrets.yaml
+++ b/charts/nextcloud/templates/secrets.yaml
@@ -17,9 +17,9 @@ data:
   {{- else }}
   nextcloud-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
-  {{- with .Values.metrics.token }}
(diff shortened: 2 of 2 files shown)


charts/nextcloud/templates/secrets.yaml
```diff
@@ -17,9 +17,9 @@ data:
...
   nextcloud-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
-  {{- with .Values.metrics.token }}
-  nextcloud-token: {{ . | b64enc | quote }}
-  {{- else }}
+  {{- if .Values.metrics.token }}
+  nextcloud-token: {{ .Values.metrics.token | b64enc | quote }}
+  {{- else if .Values.metrics.enabled }}
   nextcloud-token: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
```
jessebot, line 24:
```suggestion
  {{- if and .Values.metrics.enabled .Values.metrics.token }}
  nextcloud-token: {{ .Values.metrics.token | b64enc | quote }}
  {{- else if and .Values.metrics.enabled (not .Values.metrics.token) }}
  nextcloud-token: {{ randAlphaNum 10 | b64enc | quote }}
  {{- end }}
```

jessebot: I think the above change would ensure that we only add it if metrics are enabled in the first place?

kahirokunn: Thx!!!

jessebot (approved):
Looks good to me. Just waiting on tests to finish :)

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.7
+version: 4.7.0
```
jessebot, line 3: hmm, wait, should this be a patch or minor version? I feel like patch, right?

kahirokunn: You are certainly right.
I have incorporated that correction!
thx.

jessebot (approved):

PR #564, opened 2024-04-26
feat: Update appVersion to 29.0.0

provokateurin:
# Pull Request

Latest version.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.6
-appVersion: 28.0.4
+version: 4.6.7
+appVersion: 29.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
   - nextcloud


jessebot (approved):

provokateurin:
We had this discussions a few times already and my take is the following: The desired version can be configured by the user in the values.yaml and the appVersion is only a default value. The chart works with any Nextcloud version, so this is not a breaking change.
The same reasoning is usually applied for libraries with semver and their dependencies. If you update a dependency of that library it is only a patch bump regardless of the bump type the dependency had.

twistedgrim:
Gotta love breaking changes with minor version bumps. Tried to roll back version after the new minor version bump.. 🙃 

```
Can't start Nextcloud because the version of the data (29.0.0.19) is higher than the docker image version (28.0.4.1) and downgrading is not supported. Are you sure you have pulled the newest image version?
```

provokateurin:
Hey, please stay friendly. I see that people are frustrated by this. Do you think simply bumping the major version of the chart as well will fix it?

twistedgrim:
Not trying to not say anything unfriendly. Thank you for the work, and I understand this is community maintained.
It just feels like we might need some guidelines on what constitutes the various semver changes within the helm chart. Is it the helm chart itself or the app releases. IMO if there is a major change within the app it should constitute at least a minor change within the chart so one could investigate further if they wish. 
Again, I totally understand this is community driven and there is nothing governing the versioning requirements of the helm chart here.

provokateurin:
I think the alternative would be to drop the appVersion completely and have everyone explicitly configure the docker image version they want. That way no new release is needed for changing the default version which I'd see as a benefit because this has to happen every release right now.

PR #566, opened 2024-05-04
Allow to set environment variables to Nginx container

### LINKED ISSUE #494: Feature: Setting for timezone in nextcloud-nginx container
## Description of the change

Currently I can set the timezone for `nextcloud` container via `extraEnv` and installing `tzdata` in `postStartCommand`. However, I can't find any similar setting for `nextcloud-nginx` container in the Helm chart. This leads to situation, where `nextcloud` container is two hours "ahead" compared to `nextcloud-nginx` container.

I think the most straightforward way to add this feature would be adding similar `extraEnv` for `nextcloud-nginx` as there already is for `nextcloud` container, but I've got to admit that I'm no expert when it comes to Helm charts.

## Benefits

Troubleshooting is easier, when both containers have the same timezone.

## Additional information

I'm currently running Nextcloud with this configuration:

```yaml
image:
  repository: nextcloud
  tag: 27.1.4-fpm-alpine

lifecycle:
  postStartCommand:
    [
      "/bin/sh",
      "-c",
      "apk add tzdata",
      "&&",
      "crond",
    ]

nginx:
  ## You need to set an fpm version of the image for nextcloud if you want to use nginx!
  enabled: true
  image:
    repository: nginx
    tag: 1.21.6-alpine
```
    
I noticed that by default the timezone is UTC, but if I run manually `export TZ=Europe/Helsinki` inside `nextcloud-nginx` container, I get the correct time with `date`.

fmenabe:
# Pull Request

## Description of the change

Add the `nginx.extraEnv` parameter allowing to pass environment variables to the Nginx container. 

## Benefits

This is for example useful to set the timezone of the Nginx container through  *TZ* environment variable  (cf #494 and https://github.com/nginxinc/docker-nginx/issues/149).

## Possible drawbacks

None, the new parameter is optional.

## Applicable issues

- fixes #494 

## Checklist 

### DIFF
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -149,6 +149,10 @@ spec:
         - name: {{ .Chart.Name }}-nginx
           image: "{{ .Values.nginx.image.repository }}:{{ .Values.nginx.image.tag }}"
           imagePullPolicy: {{ .Values.nginx.image.pullPolicy }}
+          {{- with .Values.nginx.extraEnv }}
+          env:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - name: http
               protocol: TCP
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 4.6.8
+version: 4.6.9
 appVersion: 29.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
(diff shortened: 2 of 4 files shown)


charts/nextcloud/templates/deployment.yaml
```diff
@@ -149,6 +149,10 @@ spec:
         - name: {{ .Chart.Name }}-nginx
           image: "{{ .Values.nginx.image.repository }}:{{ .Values.nginx.image.tag }}"
           imagePullPolicy: {{ .Values.nginx.image.pullPolicy }}
+          {{- if .Values.nginx.extraEnv }}
+          env:
+            {{- toYaml .Values.nginx.extraEnv | nindent 12 }}
+          {{- end }}
```
jessebot, line 155:
```suggestion
          {{- with .Values.nginx.extraEnv }}
          env:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```
with is an implicit if and it reduces a little bit of code here. Also, have you tried a helm template with .Values.nginx.extraEnv set to make sure this templates correctly?

jessebot (commented):
looks ok to me, just left a really minor comment :)

charts/nextcloud/templates/deployment.yaml
fmenabe: Thanks for the review ! I applied the correction and retested a deployment (with and without nginx.extraEnvs) and it works as expected.

jessebot (approved):

PR #572, opened 2024-05-22
Update _helpers.tpl

bastikus:
Returned default values for the external database

# Pull Request

## Description of the change

Helm chart does not work when using an external database for NextCloud. "key" is not defined.

## Checklist 


provokateurin:
This seems wrong, you should simply specify the key values as they are mandatory and defaults are not going to work either because every setup is different.

bastikus:
> This seems wrong, you should simply specify the key values as they are mandatory and defaults are not going to work either because every setup is different.

Default values are indeed defined in `values.yaml` for external database, but secret for external database is not created in `db-secret.yaml`, nor are these values used in any other manifests.

provokateurin:
Hm I think having default values in the values.yaml for those fields makes no sense because everyone has to override them anyway. @jessebot what do you think?

provokateurin:
> Either way, I agree that we should not have default secret keys, as they have the potential to be wrong or unused. Would you like me to submit that PR? Do we still want to merge this one in the meantime?

Yeah a PR would be nice, it's probably the same for other configs as well. Unfortunately this will be a breaking change :/

PR #574, opened 2024-05-29
add renovatebot to run hourly in a self-hosted github action instead of using dependabot

### LINKED ISSUE #498: Automatic PRs for updating the default image version
## Description of the change

Right now we always have to manually create PRs to update the default image version to the latest one. We should automate this to automatically create the PRs. We could also enable automerge so that it gets merged if the pipeline is green. I don't think there is a possibility it could break something for us.

## Benefits

No manual PR work, faster updates to new versions.

## Possible drawbacks

None really.

## Additional information

Technically we don't need to do this as the chart is version agnostic. We could instead remove the default version and force everyone to always set the version themselves. That way we also have less "useless" releases where only the default version was changed. This would be a breaking change though so we might not want to do it.

jessebot:
# Pull Request

## Description of the change

This switches us from Dependabot (can't do helm updates) to [RenovateBot](https://github.com/renovatebot/renovate) (can do some helm updates).

## Benefits

This would generate all the same PRs that dependabot would, but it would also catch sub-chart updates, updates to the main appVerison image, and updates to the nested image versions (like for the metrics image).

## Possible drawbacks

It cannot update the `version` in Chart.yaml for docker image updates, however, it _can_ update the version for sub-chart updates. In the meantime, it will still generate a useful PR, to update what it can update, that can be edited by a maintainer to manually bump the version. There's an [ongoing discussion](https://github.com/renovatebot/renovate/discussions/28861#discussioncomment-9326722) about making this actually bump the `version` in Chart.yaml for updates to docker images and appVerison, so when that is solid, I'll port those changes here as well.

## Applicable issues

- fixes #498 

## Additional information

Before this can work, we need a nextcloud member with org settings access to create a GitHub App as described [here](https://github.com/renovatebot/github-action/blob/main/README.md#example-with-github-app):

>Example with GitHub App

> Instead of using a Personal Access Token (PAT) that is tied to a particular user you can use a [GitHub App](https://docs.github.com/en/developers/apps/building-github-apps) where permissions can be even better tuned. [Create a new app](https://docs.github.com/en/developers/apps/creating-a-github-app) and configure the app permissions and your config.js as described in the [Renovate documentation](https://docs.renovatebot.com/modules/platform/github/#running-as-a-github-app).

> Generate and download a new private key for the app, adding the contents of the downloaded .pem file to Secrets (repository settings) with the name private_key and app ID as a secret with name app_id.

## Checklist 


provokateurin:
I think it is better to use the Github App instead of Actions as it avoids potentially leaking any secrets and doesn't need any config except for the main config file. I just requested the App to be enabled on this repo, so hopefully someone with access allows it soon.

I will take a closer look at the changes of this PR soon.

provokateurin:
> The action does actually use a GitHub App to get the one-time token, but an org admin has to create it. The reason we can't use the default one from mend.io (the company that maintains the official renovateBot github app) is because it will not let us do custom post upgrade tasks. That is a self-hosted feature.

Ok I didn't know that.

> Also, it does not leak any secrets. That is not possible, even with debug logging on, in part because github automatically prunes known repo secrets

I know, but at Nextcloud we have a special problem: Many more or less trustworthy people are part of the Github organization which makes it possible for them to see the secrets in the settings. I'm not sure how problematic this token is, but this problem should be kept in mind.

provokateurin:
Ok interesting, maybe this was fixed by Github at some point. This problem was explained to me by one of our security peeps last year, so I just trusted them :see_no_evil:

PR #575, opened 2024-05-29
only run chart linting for specific files instead of trying to catch every exception

jessebot:
# Pull Request

## Description of the change

Only run the chart linting action for:
- `charts/nextcloud/Chart.yaml`
- `charts/nextcloud/values.yaml`
- `charts/nextcloud/templates/**` (everything in the templates directory)

## Benefits

This is better than trying to manually exclude all the possible things we don't want to run it on.

## Possible drawbacks

Can't think of any.

## Additional information

This would fix the issue where this PR, https://github.com/nextcloud/helm/pull/574, is trying to run the chart linter 🤦 

## Checklist 

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -23,6 +23,7 @@ jobs:
 
   lint-test:
     runs-on: ubuntu-22.04
+    needs: changes
     if: needs.changes.outputs.src != 'false'
     steps:
       - name: Checkout
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -3,12 +3,27 @@ name: Lint and Test Charts
 on:
   pull_request:
     paths:
-      - 'charts/nextcloud/Chart.yaml'
-      - 'charts/nextcloud/values.yaml'
-      - 'charts/nextcloud/templates/**'
 jobs:
+  changes:
+    runs-on: ubuntu-latest-low
+
+    outputs:
+      src: ${{ steps.changes.outputs.src}}
+
+    steps:
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
+        id: changes
+        continue-on-error: true
+        with:
+          filters: |
+            src:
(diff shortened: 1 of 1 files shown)


provokateurin:
Hm now the PR doesn't pass, because the workflow is set to be required. Could you adapt the workflow so that there is a summary job that always like in https://github.com/nextcloud/.github/blob/master/workflow-templates/node.yml? Then we can mark that one as required and we won't accidentally merge with red CI.

provokateurin (changes requested):
Hm this looks wrong, as the chart testing step should be skipped since the changes in this PR didn't touch any of those files.

provokateurin:
You can just change the name if you like. I'll change which workflow is required afterwards (but it will require every PR to be rebased).

provokateurin:
Looks like it is also fine to have the required job being skipped? Then we don't even need the summary job which was probably was just added for compatibility.

.github/workflows/lint-test.yaml
```diff
@@ -52,3 +64,17 @@ jobs:
...
+    runs-on: ubuntu-latest-low
+    if: always()
+    needs: [changes]
+
+    # This is the summary, we just avoid to rename it so that branch protection rules still match
+    name: Lint and Test Charts
+
+    steps:
+      - name: Summary status
+        run: if ${{ needs.changes.outputs.src != 'false' }}; then exit 1; fi
```
provokateurin, line 80: Just deleted this part here

.github/workflows/lint-test.yaml
```diff
@@ -2,17 +2,29 @@ name: Lint and Test Charts
...
-      - 'LICENSE'
-      - 'README.md'
-
+    paths:
 jobs:
+  changes:
+    runs-on: ubuntu-latest-low
+
+    outputs:
+      src: ${{ steps.changes.outputs.src}}
```
provokateurin, line 11: This output could be renamed, but in the end it doesn't really matter.

provokateurin (changes requested):

.github/workflows/lint-test.yaml
jessebot: I left it to save on typing

PR #580, opened 2024-06-03
Update Postgresql, MariaDB, and Redis subcharts to the latest (⚠️ Major updates)

jessebot:
# ⚠️ Major updates to sub charts

## Description of the change

- Updates postgresql chart from `12.12.10` to `15.5.0`
- Updates mariadb chart from `12.2.9` to `18.2.0`
- Updates redis chart from `17.13.*` to `19.5.0`
- Adds a new job to test postgresql installs

As always, you need to check the official upgrade instructions from both bitnami and the upstream providers. Please see the following docs before attempting the nextcloud helm chart upgrade:

- https://github.com/bitnami/charts/tree/main/bitnami/postgresql#to-1300
- https://github.com/bitnami/charts/tree/main/bitnami/mariadb#to-1400
- https://github.com/bitnami/charts/tree/main/bitnami/redis#to-1800

## Benefits

We were pretty out of date on all our subcharts, and this also grants us an end to end test with postgresql, which we've been missing for a while.

## Possible drawbacks

- I don't have a test for mysql, only because I don't actually use mariadb, and don't have the time to create one right now, but if others want to submit that test, I'd be open to testing it locally :)
- users will need to still follow the major upgrade docs listed above before they can upgrade, as is the case for all major upgrades of redis, mariadb, and postgresql
- I have hardcoded a password for the testing of postgresql. We can easily update this to be a repo secret and then no password will be logged, but that requires someone with better access than myself to add that repo secret, and then tell me know what the name of it is, so I can update the `ct` command.

## Applicable issues

I feel like we had an issue for postgresql testing, but I can't find it right now 🤔 

## Additional information

You can see an example of the new testing workflow here where I ran it on my personal repo for this feature branch:
https://github.com/jessebot/nextcloud-helm/actions/runs/9347132143
Screenshot of the summary section for the jessebot/nextcloud-helm feature branch workflow showing 4 jobs. The first job is the changes job, which then feeds into the linting job, which then triggers both the internal-database test and postgresql-database test. All 4 show as green, meaning success.

In the future, renovateBot will take care of these updates for us, and then we just need to review them. I'm doing this manually now, because we're so very behind.

## Checklist 

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -5,7 +5,7 @@ on:
     paths:
 jobs:
   changes:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest-low
 
     outputs:
       src: ${{ steps.changes.outputs.src}}
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -73,11 +73,21 @@ jobs:
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.6.1
 
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --target-branch ${{ github.event.repository.default_branch }})
+          if [[ -n "$changed" ]]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Create kind cluster
         uses: helm/kind-action@v1.10.0
+        if: steps.list-changed.outputs.changed == 'true'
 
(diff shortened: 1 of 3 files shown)


provokateurin (approved):
I didn't test, but changes LGTM

provokateurin:
Ah yeah now the required workflows are messing this up again. Since everyone has to rebase anyway due to the chart version conflicts it's fine to just set the new jobs to be required.

provokateurin:
@jessebot feel free to merge, I already updated the required jobs.

PR #581, opened 2024-06-09
feat(nextcloud): add notify_push support

### LINKED ISSUE #109: High Performance File Backend

wrenix:
# Pull Request

## Description of the change
Not yet tested

## Benefits
- support notify_push
- improve ServiceMonitor (to collect data from nextcloud-exporter and notify_push)

## Applicable issues

- fixes #109 
- closes #551

## Checklist 

TODO
  - [x] redis password from existing secret (or URL)
  - [x] put redis password from env to secrets
  - [x] solve bootstrap problem (nextcloud and notify_push needs to be online)
  - [x] option to set urlEnv on notifyPush (for external redis)

Todo after Review:
  - [x] redis-env https://github.com/nextcloud/helm/pull/581#discussion_r1903146252
  - [ ] selftest as unittest: https://github.com/nextcloud/helm/pull/581#discussion_r1903146391

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -83,6 +83,10 @@ jobs:
           - name: Horizontal Pod Autoscaling Enabled
             helm_args: '--helm-extra-set-args "--values charts/nextcloud/test-values/hpa.yaml"'
 
+          # test the helm chart with notify push enabled
+          - name: Notify Push Enabled
+            helm_args: '--helm-extra-set-args "--values charts/nextcloud/test-values/notify_push.yaml"'
+
           # test the helm chart with s3 as the primary storage
           - name: S3 Enabled as Primary Storage
             # we need to skip the clean up so we can test adding a file
--- a/charts/nextcloud/files/notify_push.sh.tpl
+++ b/charts/nextcloud/files/notify_push.sh.tpl
@@ -1,0 +1,8 @@
+#!/bin/sh
+/var/www/html/occ app:enable notify_push
+/var/www/html/occ config:app:set notify_push base_endpoint --value="http{{ if .Values.ingress.tls }}s{{ end }}://{{ .Values.nextcloud.host }}{{ .Values.notifyPush.ingress.path }}"
+{{/*
+The command "setup" runs a check, which need a running nextcloud (but we try to configurate it during startup).
+So that command always failure and we stuck in bootloop.
+/var/www/html/occ notify_push:setup "http{{ if .Values.ingress.tls }}s{{ end }}://{{ .Values.nextcloud.host }}{{ .Values.notifyPush.ingress.path }}"
+*/}}
--- a/charts/nextcloud/templates/notify_push/deployment.yaml
+++ b/charts/nextcloud/templates/notify_push/deployment.yaml
@@ -1,0 +1,65 @@
+{{- if .Values.notifyPush.enabled }}
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "nextcloud.fullname" . }}-notify-push
+  labels:
+    {{- include "nextcloud.labels"  ( dict "component" "notify-push"  "rootContext" $ ) | nindent 4 }}
+spec:
+  replicas: {{ .Values.notifyPush.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "nextcloud.selectorLabels"  ( dict "component" "notify-push"  "rootContext" $ ) | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        {{- toYaml .Values.notifyPush.podAnnotations | nindent 8 }}
+      labels:
+        {{- include "nextcloud.selectorLabels"  ( dict "component" "notify-push"  "rootContext" $ ) | nindent 8 }}
+        {{- if .Values.redis.enabled }}
+        {{ template "nextcloud.redis.fullname" . }}-client: "true"
+        {{- end }}
+        {{- with .Values.notifyPush.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- with .Values.notifyPush.image.pullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end}}
+      {{- end }}
+      containers:
+        - name: notify-push
+          {{- with .Values.notifyPush.image }}
+          image: "{{ .registry }}/{{ .repository }}:{{ .tag }}"
+          imagePullPolicy: {{ .pullPolicy }}
+          {{- end }}
+          env:
+            - name: PORT
+              value: "7867"
+            - name: METRICS_PORT
+              value: "9867"
+            {{- include "nextcloud.env.database" . | nindent 12 }}
+            {{- if .Values.redis.enabled }}
+            {{- include "nextcloud.env.redis" . | nindent 12 }}
+            {{- end }}
+            - name: NEXTCLOUD_URL # deployment.namespace.svc.cluster.local
+              value: "http{{ if .Values.notifyPush.https }}s{{ end }}://{{ template "nextcloud.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
+            {{- with .Values.notifyPush.extraEnv }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          ports:
+            - name: http
+              containerPort: 7867
+            - name: metrics
+              containerPort: 9867
+          {{- with .Values.notifyPush.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          securityContext:
+            runAsUser: 1000
+            runAsNonRoot: true
+{{- end }}
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -466,11 +466,14 @@ externalDatabase:
   ## Use a existing secret
   existingSecret:
     enabled: false
-    # secretName: nameofsecret
+    # -- e.g. nameofsecret
+    secretName:
     usernameKey: db-username
     passwordKey: db-password
-    # hostKey: db-hostname-or-ip
-    # databaseKey: db-name
+    # -- e.g. db-hostname-or-ip
+    hostKey:
+    # -- e.g. db-name
+    databaseKey:
 
 ##
 ## MariaDB chart configuration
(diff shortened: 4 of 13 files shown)


AndreKoepke:
Can we ensure that the notify_push-plugin is installed? Maybe something like this?
```yaml
 lifecycle:
      postStart:
        exec:
          command: ["occ",  "app:install notify_push"]
```

And we have to active that plugin by running `sudo -u www-data ./occ notify_push:setup https://NEXTCLOUD_HOST/push`. Maybe we should add a little script like this (pseudocode):

```text
if (!notify_push installed)
  install notify_push
/occ notify_push:setup https://NEXTCLOUD_HOST/push
```

provokateurin:
I think it makes sense to give this option so the installation and setup is completely automatic, but I'd rather put it behind a second config flag:
```yaml
notify_push:
	enabled: true
	automatic_setup: true
```
Maybe some people don't want to have this done automatically, so it's nice to give them the option.

wrenix:
Therefore we has that hook script, i write a ConfigMap for it

AndreKoepke:
```
andre@server:~/k8s/nextcloud$ helm upgrade -n nextcloud akops-nextcloud -f values.yml ./helm/charts/nextcloud/
Error: UPGRADE FAILED: template: nextcloud/templates/notify_push/deployment.yaml:41:31: executing "nextcloud/templates/notify_push/deployment.yaml" at <$.Values.global.image.registry>: nil pointer evaluating interface {}.image
```

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+        - name: {{ . }}
+        {{- end}}
+      {{- end }}
+      containers:
+        - name: notify-push
+          {{- with .Values.notifyPush.image }}
+          image: "{{ coalesce $.Values.global.image.registry .registry }}/{{ .repository }}:{{ .tag }}"
+          imagePullPolicy: {{ coalesce $.Values.global.image.pullPolicy .pullPolicy }}
+          {{- end }}
+          imagePullPolicy: {{ .Values.metrics.image.pullPolicy }}
```
AndreKoepke, line 44: Doubled `imagePullPolicy`

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,71 @@
...
+      {{- with .Values.notifyPush.image.pullSecrets }}
+      imagePullSecrets:
+        {{- range . }}
+        - name: {{ . }}
+        {{- end}}
+      {{- end }}
+      containers:
+        - name: notify-push
+          {{- with .Values.notifyPush.image }}
+          image: "{{ coalesce $.Values.global.image.registry .registry }}/{{ .repository }}:{{ .tag }}"
```
AndreKoepke, line 41: I can't find a reference of `$.Values.global.image.registry`

AndreKoepke (changes requested):
When I remove the global-references, then I got this error:

```console
andre@server:~/k8s/nextcloud$ helm upgrade -n nextcloud akops-nextcloud -f values.yml ./helm/charts/nextcloud/
Error: UPGRADE FAILED: YAML parse error on nextcloud/templates/notify_push/deployment.yaml: error converting YAML to JSON: yaml: line 41: did not find expected key
```

wrenix:
Oh sorry, that was a copy-paste error.
normally i has on my helm-charts a `global.image` part to overwrite for registry and so on.

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+              value: "7867"
+            - name: METRICS_PORT
+              value: "9867"
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-db" .Release.Name) }}
+                  key: {{ .Values.externalDatabase.existingSecret.databaseURLKey }}
+            - name: REDIS_URL
+              value: "redis://{{ template "nextcloud.redis.fullname" . }}-master:{{ .Values.redis.master.service.ports.redis | quote }}"
```
AndreKoepke, line 55: This quote causing trouble. Without `| quote` I was able to generate the template.

AndreKoepke:
```suggestion
              value: "redis://{{ template "nextcloud.redis.fullname" . }}-master:{{ .Values.redis.master.service.ports.redis }}"
```

AndreKoepke:
I was able to try an install. The result was this:

```
efaulted container "nextcloud" out of: nextcloud, nextcloud-nginx, nextcloud-cron
Configuring Redis as session handler
=> Searching for scripts (*.sh) to run, located in the folder: /docker-entrypoint-hooks.d/before-starting
==> Running the script (cwd: /var/www/html): "/docker-entrypoint-hooks.d/before-starting/notify_push.sh"
notify_push already installed
✓ redis is configured
🗴 push server is not receiving redis messages (received 272721789, got 0)
==> Failed at executing "/docker-entrypoint-hooks.d/before-starting/notify_push.sh". Exit code: 1
```

wrenix:
okay, redis password should work:
  - [ ] but not on existing secrets TODO
We has to reorganize the secrets.

PS: found that the redis password is not stored by default in a secret ....

--- 

oh yes, that is a bootstrap problem. maybe we could remove it on one part.

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,70 @@
...
+                secretKeyRef:
+                  name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-db" .Release.Name) }}
+                  key: {{ .Values.externalDatabase.existingSecret.databaseURLKey }}
+            - name: REDIS_URL
+              value: "redis://{{ if .Values.redis.auth.enabled }}:{{ .Values.redis.auth.password }}@{{ end }}{{ template "nextcloud.redis.fullname" . }}-master:{{ .Values.redis.master.service.ports.redis }}"
+            - name: NEXTCLOUD_URL # deployment.namespace.svc.cluster.local
+              value: "http{{ if .Values.notifyPush.https }}s{{ end }}://{{ template "nextcloud.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
+          ports:
+            - name: http
+              containerPort: 7867
```
AndreKoepke, line 60:
```suggestion
              containerPort: 80
```

AndreKoepke:
With the fixed port, I still unable to run it.

Logs from notify_push pod:
```
[2024-06-19 06:44:36.746890 +00:00] ERROR [notify_push] src/main.rs:78: Self test failed: Error while communicating with nextcloud instance: error sending request for url (http://akops-nextcloud.nextcloud.svc.cluster.local:8080/index.php/apps/notify_push/test/version): error trying to connect: tcp connect error: Connection refused (os error 111)
```

charts/nextcloud/templates/notify_push/deployment.yaml
wrenix: Overseen this comment and suggestion but I found it either ;)

charts/nextcloud/templates/notify_push/deployment.yaml
wrenix: Hmm, that is strange - I do like to run pods with less priority and permission (port over 1000) and reNAT it with the service to an port like 80 ...


That is the reason why I set the ENV and this port to 7867.
So I like to debug, why the ENV does not work correct.

wrenix: Hmm also notify_push mean, this is the default port:
https://github.com/nextcloud/notify_push?tab=readme-ov-file#configuration

So, how should it work this way 🤔

AndreKoepke: I got a connection timeout when I started it and only saw the port 80 in the service. I was a bit in hurry.

Your right, this should be the correct port-setting.

.github/workflows/lint-test.yaml
```diff
@@ -81,6 +81,10 @@ jobs:
           # test the helm chart with horizontal pod autoscaling enabled
           - name: Horizontal Pod Autoscaling Enabled
             helm_args: '--helm-extra-set-args "--set=hpa.enabled=true --set=hpa.minPods=2 --set=hpa.maxPods=3 --set=hpa.targetCPUUtilizationPercentage=75"'
+          
+          # test the helm chart with notify push enabled
+          - name: Notify Push Enabled
+            helm_args: '--helm-extra-set-args "--set=notifyPush.enabled=true"'
```
jessebot, line 87: Feel free to remove or tweak this @wrenix 🙏

charts/nextcloud/templates/notify_push/deployment.yaml
Sn3akyF0x, line 1: Hello,

First of all, thank you for your amazing job!
I'm trying to deploy Nextcloud with an external Redis-like deployment (keydb).
I've found some problems in configuring this for notify_push, because there  are, in my opinion, some wrongly hardcoded values. Here what's inside the template:

```
- name: REDIS_URL
  value: "redis://{{ if .Values.redis.auth.enabled }}:{{ .Values.redis.auth.password }}@{{ end }}{{ template "nextcloud.redis.fullname" . }}-master:{{ .Values.redis.master.service.ports.redis }}"
```

This will force me to have a Redis deployment with a "[...]-master" name, moreover it takes for granted that the user wants to deploy the provided Redis subchart.

Sn3akyF0x: Thank you Wrenix for the clarification, I'm looking forward to your improvements to be merged!

AndreKoepke:
@wrenix cool changes. I tried your branch and it worked directly without any issue in my environment. :)

AndreKoepke (approved):

charts/nextcloud/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+This Helm-Chart use semantic-releases, so only major version contains breaking changes.
+
+Here we list, what is changed.
+
+
+## v7
+
+- move `metrics.serviceMonitor` to `prometheus.serviceMonitor`: It us used for nextcloud-exporter and notify-push
```
provokateurin, line 10:
```suggestion
- move `metrics.serviceMonitor` to `prometheus.serviceMonitor`: It is used for nextcloud-exporter and notify-push
```

charts/nextcloud/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+This Helm-Chart use semantic-releases, so only major version contains breaking changes.
```
provokateurin, line 3: Or when a new major Nextcloud version is used.
It's not a breaking change because it only changes the default, but we have learned that it causes trouble for many users, so we bump the major version anyway.

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+              value: "7867"
+            - name: METRICS_PORT
+              value: "9867"
+            {{- include "nextcloud.env.database" . | nindent 12 }}
+            {{- with .Values.notifyPush.redisURLEnv }}
+            - name: "REDIS_URL"
+              {{- toYaml . | nindent 14 }}
+            {{- else }}
+            {{- if not .Values.redis.enabled }}
+            {{- fail "notify-push needs an redis (redis.enabled=true or notifyPush.redisURLEnv={...})" }}
```
provokateurin, line 55: Good to have it here, but it would be nice to mention this in the readme as well.

charts/nextcloud/files/notify_push.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+/var/www/html/occ app:install notify_push
+/var/www/html/occ config:app:set notify_push base_endpoint --value="https://{{ .Values.nextcloud.host }}{{ .Values.notifyPush.ingress.path }}"
```
provokateurin, line 3: Hardcoded to https? This seems wrong to me.
Also accessing it through the external host shouldn't be necessary and wouldn't work if the port is different.

charts/nextcloud/values.yaml
```diff
@@ -675,12 +683,50 @@ tolerations: []
...
+  # -- enable another deployment to handle notify_push (sometimes called ClientPush)
+  enabled: false
+  autoSetup: false
+
+  replicaCount: 1
+
+  image:
+    registry: docker.io
+    # -- image of notify_push (there is no official image yet: https://github.com/nextcloud/notify_push/issues/106)
+    repository: miles170/notify_push
```
provokateurin, line 706: This doesn't feel good to me, anything that is not at least semi-official (like the nextcloud docker image) is a risky dependency to add.

provokateurin (changes requested):
This looks mostly good to me, but I have some reservations.

You should also add some documentation to the readme and update the values tables.

@jessebot can you also take a look?

charts/nextcloud/files/notify_push.sh.tpl
provokateurin: Or alternatively the hostname in the container should be bent to localhost similar to https://github.com/NixOS/nixpkgs/blob/a23b0bdc68f56700d03aec9598eeeae6975f0ade/nixos/modules/services/web-apps/nextcloud-notify_push.nix#L138-L141

charts/nextcloud/CHANGELOG.md
provokateurin: Yes

charts/nextcloud/values.yaml
provokateurin: Could you try the AIO container? I'd be more comfortable with that, but if it doesn't work we can use this image for now.

charts/nextcloud/files/notify_push.sh.tpl
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+/var/www/html/occ app:install notify_push
```
provokateurin, line 2:
```suggestion
/var/www/html/occ app:enable notify_push
```
This will both install and enable the app. I'd expect that the app is automatically enabled, if it was disabled but installed previously.

charts/nextcloud/files/notify_push.sh.tpl
```diff
@@ -0,0 +1,4 @@
...
+/var/www/html/occ config:app:set notify_push base_endpoint --value="http{{ if .Values.ingress.tls }}s{{ end }}://{{ .Values.nextcloud.host }}{{ .Values.notifyPush.ingress.path }}"
+# /var/www/html/occ notify_push:setup "http{{ if .Values.ingress.tls }}s{{ end }}://{{ .Values.nextcloud.host }}{{ .Values.notifyPush.ingress.path }}"
```
provokateurin, line 4: Why is this commented out?
Does the command not work due to startup order creating a circular dependency?
Either way I don't think a commented out command is helpful, so please remove one of them.

charts/nextcloud/templates/metrics/service.yaml
```diff
@@ -1,35 +1,38 @@
...
   loadBalancerIP: {{ . }}
   {{- end }}
   {{- end }}
+  selector:
+    app.kubernetes.io/name: {{ include "nextcloud.name" $ }}
+    app.kubernetes.io/instance: {{ $.Release.Name }}
+    app.kubernetes.io/component: metrics
   ports:
     - name: metrics
-      port: 9205
```
provokateurin, line 29: Why is the port changed? This could be an unintended and unnecessary breaking change for some setups.

charts/nextcloud/templates/notify_push/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+          {{- end }}
+          env:
+            - name: PORT
+              value: "7867"
+            - name: METRICS_PORT
+              value: "9867"
+            {{- include "nextcloud.env.database" . | nindent 12 }}
+            {{- with .Values.notifyPush.redisURLEnv }}
+            - name: "REDIS_URL"
+              {{- toYaml . | nindent 14 }}
```
provokateurin, line 52: This is a bit ugly, a generic option like we already have for the server deployment would be nicer: https://github.com/nextcloud/helm/blob/70403bcdc8a98af6d9ab17110f4c1682c3a1ec9e/charts/nextcloud/values.yaml#L259-L264

.github/workflows/lint-test.yaml
```diff
@@ -83,6 +83,10 @@ jobs:
           - name: Horizontal Pod Autoscaling Enabled
             helm_args: '--helm-extra-set-args "--values charts/nextcloud/test-values/hpa.yaml"'
 
+          # test the helm chart with notify push enabled
+          - name: Notify Push Enabled
+            helm_args: '--helm-extra-set-args "--values charts/nextcloud/test-values/notify_push.yaml"'
```
provokateurin, line 88: So we only test that the deployment succeeds, could we use the notify_push self-test check (or whatever it is called) to ensure it works correctly?
Or would that already be covered automatically because the deployments won't succeed otherwise?

charts/nextcloud/files/notify_push.sh.tpl
provokateurin: Yes please

charts/nextcloud/templates/metrics/service.yaml
provokateurin: Ok fine, just add it to the changelog to avoid surprises.

cmdPromptCritical:
any help needed? happy to support.

PR #596, opened 2024-07-23
only set `spec.replicas` in Nextcloud Deployment if `.Values.hpa.enabled` is set to `false`

### LINKED ISSUE #561: HPA kills new pods instantly after creation
## Describe your Issue

When I activate HPA in the helmet chart, a pod is initially scheduled (which is correct). As soon as I synchronize files, the CPU load of this pod increases to >60%. So HPA tries to schedule new pods. These are then also scheduled, but killed again immediately after container creation. This means that there is never more than one pod running at the same time, although there should be 5.

### Logs and Errors

There are no errors in the logs. The termination seems to be caused by Kubernetes itself.

## Describe your Environment

- Kubernetes distribution: k3s

- Helm Version (or App that manages helm): ArgoCD version v2.10.7+b060053

- Helm Chart Version: 4.6.6

- `values.yaml`:

```yaml
nextcloud:
  host: 10.3.28.0
  configs:
    custom.config.php: |
      <?php
        $CONFIG = array(
          "check_data_directory_permissions"=> false, # fix data directory permissions error
          "trusted_domains" => array (
            $_ENV["NEXTCLOUD_TRUSTED_DOMAINS"], # fix probes 400 error
          ),
          'trusted_proxies' => array(
            0 => '127.0.0.1',
            1 => '10.0.0.0/8',
          ),
          "forwarded_for_headers" => array("HTTP_X_FORWARDED_FOR"),
        );
  containerPort: 8080
  extraVolumes:
    - name: nginx-cache
      emptyDir: { }
  extraVolumeMounts:
    - name: nginx-cache
      mountPath: "/var/cache/nginx" # fix permission denied error
  securityContext:
    runAsUser: 901000
    runAsGroup: 901000
    runAsNonRoot: true
  podSecurityContext:
    runAsUser: 901000
    runAsGroup: 901000
    runAsNonRoot: true
service:
  type: LoadBalancer
internalDatabase:
  enabled: false
image:
  flavor: fpm
nginx:
  enabled: true
  image:
    repository: nginxinc/nginx-unprivileged
    tag: 1.25 # https://hub.docker.com/r/nginxinc/nginx-unprivileged/tags
  containerPort: 8080
  resources:
    limits:
      cpu: 200m
      memory: 128Mi
    requests:
      cpu: 100m
      memory: 64Mi
  securityContext:
    runAsUser: 901000
    runAsGroup: 901000
    runAsNonRoot: true
externalDatabase:
  enabled: true
  type: postgresql
  host: nextcloud-postgresql-primary
  database: nextcloud
  user: nextcloud
  password: nextcloud
hpa:
  enabled: true
  minPods: 1
  maxPods: 5
resources:
  limits:
    cpu: 500m
    memory: 512Mi
  requests:
    cpu: 250m
    memory: 256Mi
persistence:
  enabled: true
  existingClaim: pvc-k8s-nextcloud-app
  nextcloudData:
    enabled: true
    existingClaim: pvc-k8s-nextcloud-data
livenessProbe:
  enabled: true
readinessProbe:
  enabled: true
startupProbe:
  enabled: true
```

## Additional context, if any
I found a possible solution for this issue. As mentioned in [this StackOverflow article](https://stackoverflow.com/questions/72840685/horizontalpodautoscaler-scales-up-pods-but-then-terminates-them-instantly), the `replicas` parameter cannot be used in the deployment resource if a HPA definition is used.

I am using NFS as persistant storage with the [NFS CSI driver](https://github.com/kubernetes-csi/csi-driver-nfs). The PVC has RWX access mode.

jessebot:
## Description of the change

- Sets`spec.replicas` in Nextcloud Deployment _only if_ `hpa.enabled` in values.yaml is set to `false`.
- update hpa ci test to use `hpa.minPods=2`

## Benefits

This _should_ unbreak the environments that use `hpa.enabled` in their values.yaml that also use Argo CD?

## Possible drawbacks

None that I can think of after #598 was merged! Open to discussion on this one as always :)

## Applicable issues

- fixes #561

## Checklist 

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -80,7 +80,7 @@ jobs:
 
           # test the helm chart with horizontal pod autoscaling enabled
           - name: Horizontal Pod Autoscaling Enabled
-            helm_args: '--helm-extra-set-args "--set=hpa.enabled=true --set=hpa.minPods=1 --set=hpa.maxPods=3 --set=hpa.targetCPUUtilizationPercentage=75"'
+            helm_args: '--helm-extra-set-args "--set=hpa.enabled=true --set=hpa.minPods=2 --set=hpa.maxPods=3 --set=hpa.targetCPUUtilizationPercentage=75"'
 
     steps:
       - name: Checkout
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 5.3.0
+version: 5.3.1
 appVersion: 29.0.4
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -203,7 +203,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `startupProbe.timeoutSeconds`                              | When the probe times out                                                                            | `5`                        |
 | `startupProbe.failureThreshold`                            | Minimum consecutive failures for the probe                                                          | `30`                       |
 | `startupProbe.successThreshold`                            | Minimum consecutive successes for the probe                                                         | `1`                        |
-| `hpa.enabled`                                              | Boolean to create a HorizontalPodAutoscaler                                                         | `false`                    |
+| `hpa.enabled`                                              | Boolean to create a HorizontalPodAutoscaler. If set to `true`, ignores `replicaCount`.              | `false`                    |
 | `hpa.cputhreshold`                                         | CPU threshold percent for the HorizontalPodAutoscale                                                | `60`                       |
 | `hpa.minPods`                                              | Min. pods for the Nextcloud HorizontalPodAutoscaler                                                 | `1`                        |
 | `hpa.maxPods`                                              | Max. pods for the Nextcloud HorizontalPodAutoscaler                                                 | `10`                       |
--- a/charts/nextcloud/templates/deployment.yaml
+++ b/charts/nextcloud/templates/deployment.yaml
@@ -17,7 +17,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if not .Values.hpa.enabled }}
   replicas: {{ .Values.replicaCount }}
+  {{- end }}
   strategy:
     {{- toYaml .Values.nextcloud.strategy | nindent 4 }}
   selector:


provokateurin:
Some kind of test would be nice, the change makes sense but I can not validate it.

.github/workflows/lint-test.yaml
```diff
@@ -129,9 +131,50 @@ jobs:
...
         id: install
         if: steps.list-changed.outputs.changed == 'true'
         run: |
             ct install --target-branch ${{ github.event.repository.default_branch }} \
             --helm-extra-set-args "--set=postgresql.enabled=true --set=postgresql.global.postgresql.auth.password=testing123456 --set=internalDatabase.enabled=false --set=externalDatabase.enabled=True --set=externalDatabase.type=postgresql --set=externalDatabase.password=testing123456"
+
+  test-hpa:
+    runs-on: ubuntu-22.04
+    needs: [changes, lint]
+    if: needs.changes.outputs.src != 'false' && github.event.label.name == 'hpa'
```
jessebot, line 144: I've set this to only run on the `hpa` PR label, so reduce need to run this check every PR :)

jessebot: It was easier than I thought, so I did this in reverse order of checking the ci first and then testing locally.

provokateurin: I think we should run it all the time, otherwise the label might be missed on a PR and then a bug can slip in unnoticed. Running one extra job is really no concern to me, especially since it doesn't take a long time to complete.

jessebot: Sure, I can make that change in just a bit :) I was considering submitting a PR for adding nginx tests as well. Would you like those to run every time as well, or only on label of `nginx container`?

provokateurin: Always would be best imo :)

jessebot: Can do! :)

provokateurin (approved):

provokateurin:
I think we need a summary job, otherwise this is not going to work when the tests are skipped because there were no changes.

charts/nextcloud/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 5.3.0
+version: 5.4.0
```
provokateurin, line 3: Only a bug fix, so 5.3.1?

provokateurin (commented):
I didn't want to drag out the other PR even further, but I think the test should set hpa.minPods to 2 to ensure it works with multiple pods. I'd assume in the testing only 1 pod gets created which doesn't test the feature well.

charts/nextcloud/Chart.yaml
jessebot: oh yeah, I guess it is a bug fix and not a feature, you're right! I'll get this fixed :)

provokateurin (approved):
HPA works :tada:

Issue #103, opened 2021-03-30
Allows jumps of multiple versions


kiriakos:
Hi, I noticed that NextCloud still cannot migrate over more than one major version, resulting in an unworkable and unrecoverable system. In order to help others avoid this tearful realization how about marking the `image:` section respectively, warning users to upgrade one version at a time?

Cheers

ftab:
For anyone who forgot this, like me (goofed and tried to run 21.0.2 over 19.0.12), and then set the image tag back a version but then Nextcloud says `Can't start Nextcloud because the version of the data (21.0.2.1) is higher than the docker image version (20.0.10.1) and downgrading is not supported. Are you sure you have pulled the newest image version?`

The quick and dirty fix is delete /var/www/html/version.php, restart the container (via kubectl delete po) to recreate version.php with the current version info, and you will see by the logs, it will think it needs to install a new Nextcloud. Restart it again and you can proceed with the upgrade.

```sh
export NEXTCLOUD_POD_NAME=`/usr/local/bin/kubectl get po -l "app.kubernetes.io/instance=nextcloud" -o jsonpath='{.items[0].metadata.name}'`
kubectl exec -it $NEXTCLOUD_POD_NAME -- bash
# inside the nextcloud pod bash:
su www-data -s /bin/bash
./occ upgrade
./occ maintenance:mode --off
```

I am using the mariadb installed by the chart. YMMV

provokateurin:
This is not possible with Nextcloud in general: https://docs.nextcloud.com/server/latest/admin_manual/maintenance/upgrade.html#approaching-upgrades

PR #602, opened 2024-07-25
Update CONTRIBUTING.md - fix #20 by adding a section on version upgrades

### LINKED ISSUE #20: Direct upgrade from 17.0 to 19.0 not supported, breaks existing installations
Hi,

the latest commit https://github.com/nextcloud/helm/commit/5c9f27e52c169cfc4f1b4e49cbc1955683d41ad2 upgrades Nextcloud from version 17 to 19. However, a direct upgrade is not supported. Users need to go to 18.0 first.

Therefore, upgrading an existing release of this chart (with default parameters, i.e., Nextcloud 17) will break on upgrade.
There should be documentation that one needs to go to 18.0 first and that this is a manual step.

Workaround after it happened: fix the version number from 19.x to 17.0.9 in `html/version.php` (which is on the persistent volume). If 19.x is mentioned there, the 18.x upgrade will abort and not even see that the database is still 17.x.

jessebot:
## Description of the change

Adds a section about upgrades to clarify we should only do one major version at a time.

## Benefits

just a little doc change to prevent gotchya :)

## Possible drawbacks

open to feedback but I think this is minor enough that it's fine.

## Applicable issues

- fixes #20

## Checklist 


provokateurin (commented):
Upgrading to the latest minor version before upgrading to the next major version should probably also be documented: https://docs.nextcloud.com/server/latest/admin_manual/maintenance/upgrade.html#approaching-upgrades
This is not only for creating PRs to update the default version but also for the people who deploy the updates.

provokateurin:
> I could improve this PR by also making a note in the values.yaml for the image.tag parameter as mentioned in https://github.com/nextcloud/helm/issues/103 if you agree that's a good idea?

Yeah that would be nice too, just to prevent people ending up in broken states...

provokateurin:
Yes that sounds wrong indeed, you can fix it in https://github.com/nextcloud/documentation

PR #613, opened 2024-07-29
Add FORWARDED_FOR_HEADERS to the reverse-proxy config

dgiebert:
## Description of the change

With the new proxy config added, I want to be able to remove all manual config in the helm chart and use env vars

## Checklist 


provokateurin:
We're just using the config from https://github.com/nextcloud/docker/blob/master/.config/reverse-proxy.config.php and try to stay in sync with it, so I'd prefer if it was submitted there first since the maintainers of the docker image will know if this makes sense. After that we can just add it here by updating the file from upstream. @jessebot what do you think about this process?

dgiebert:
Then I will go ahead and create a PR against nextcloud/docker and reference it here.

PR #623, opened 2024-08-23
Feature: Add Collabora as a subchart #611

### LINKED ISSUE #611: Feature: Add Collabora as a subchart
## Description of the change

Let's add support for [Collabora](https://www.collaboraonline.com/) as a subchart. Here's their [helm chart](https://github.com/CollaboraOnline/online/tree/master/kubernetes/helm/collabora-online) and here's their [docs](https://sdk.collaboraonline.com/docs/installation/Kubernetes.html).

- Split from #214

## Benefits

Allow users to have a document server deployed in this stack.

## Possible drawbacks

- Another thing to support.
- Unknown: Not sure how much we can facilitate actually configuring the connection between Nextcloud and Collabora 🤷

Happy to discuss anything further!

## Tasks

These are the tasks that would need to be done for this issue to be considered complete:

- [ ] add tests in ci and/or helm unit-tests
- [ ] add Collabora as a configurable sub-chart via parameters in values.yaml
- [ ] install the app is there is one?

We can add more tasks if needed.

kriegalex:
## Description of the change

- Add a dependency to the 1.1.20 collabora chart from their official repository
- Add the subchart config to values.yaml
- Add the configuration options documentation to README

## Benefits

- Install collabora as a subchart of nextcloud via helm

## Possible drawbacks

- People may try to enable it without knowing what is is and how to configure it (mainly `extra_params`, `server_name`, `aliasgroups` and `annotations`.

## Applicable issues

- fixes #611 

## Additional information

Tested on my local kubernetes cluster with a HTTPS nginx ingress. Nextcloud can communicate with Collabora. Not tested with HAProxy (though it is only a change in annotations specific to HAProxy).

## Checklist 

- [X] Parameters are documented in the README.md

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -472,6 +472,54 @@ redis:
     persistence:
       enabled: true
 
+##
+## Collabora chart configuration
+## for more options see https://github.com/CollaboraOnline/online/tree/master/kubernetes/helm/collabora-online
+##
+collabora:
+  enabled: false
+
+  # please check collabora README.md first
+  autoscaling:
+    enabled: false
+
+  collabora:
+    ## HTTPS nextcloud domain, if needed
+    # aliasgroups:
+    #   - host: "https://nextcloud.domain:443"
+    # extra_params: --o:ssl.enable=false --o:ssl.termination=true
+
+    ## Specify server_name when the hostname is not reachable directly for example behind reverse-proxy
+    # server_name: collabora.domain
+
+    # setup admin login credentials
+    existingSecret:
+      enabled: false
+      secretName: ""
+      usernameKey: "username"
+      passwordKey: "password"
+    password: examplepass
+    username: admin
+  
+  # setup ingress
+  ingress:
+    enabled: false
+    className: ""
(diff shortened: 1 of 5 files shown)


jessebot (changes requested):
looks great, but we need to ensure the tests still work, so I left a comment about how to update the GHA workflow. If you can, it would also be neat to have a test for collabora itself as well, but we can also hold off on that for a separate PR if you'd like. 🙏

charts/nextcloud/values.yaml
```diff
@@ -472,6 +472,54 @@ redis:
...
+
+    # setup admin login credentials
+    existingSecret:
+      enabled: false
+      secretName: ""
+      usernameKey: "username"
+      passwordKey: "password"
+    password: examplepass
+    username: admin
+
```
jessebot, line 503:
```suggestion

```

kriegalex:
Thanks for the change on lint tests. I'm still learning k8s, didn't think of that.

jessebot (approved):

provokateurin:
Publishing the release failed: https://github.com/nextcloud/helm/actions/runs/12341432621/job/34440050803

PR #629, opened 2024-09-19
update: 30.0.0

terop:
## Description of the change

Take latest version in use. Chart major version changed because of the updated major version of Nextcloud.

## Benefits

Users get latest features and fixes.

## Possible drawbacks

There may be bugs and / or regressions.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 5.5.6
-appVersion: 29.0.6
+version: 6.0.0
+appVersion: 30.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:
   - nextcloud


charts/nextcloud/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: nextcloud
-version: 5.5.6
-appVersion: 29.0.6
+version: 5.6.0
+appVersion: 30.0.0
```
provokateurin, line 4:
```suggestion
version: 6.0.0
appVersion: 30.0.0
```
Last time a lot of people complained we didn't make a breaking change, so let's make a breaking change for this major update.

provokateurin (changes requested):

charts/nextcloud/Chart.yaml
terop: Changed as suggested.

provokateurin (approved):
Awesome, thank you!

twistedgrim:
Thank you guys for making it a major version change. This helped me know to look closer for the major version change for nextcloud <3 It really helps for ensuring DB migrations etc work properly and we take a closer look as we use gitops automation.

PR #630, opened 2024-09-19
fix(nginx): add option to easy manage HEADERS (make it possible to set HSTS)

wrenix:
## Applicable issues

- fixes #

## Checklist 

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -331,6 +331,24 @@ nginx:
   config:
     # This generates the default nginx config as per the nextcloud documentation
     default: true
+    headers:
+      # -- HSTS settings
+      # WARNING: Only add the preload option once you read about
+      # the consequences in https://hstspreload.org/. This option
+      # will add the domain to a hardcoded list that is shipped
+      # in all major browsers and getting removed from this list
+      # could take several months.
+      # Example:
+      # "Strict-Transport-Security": "max-age=15768000; includeSubDomains; preload;"
+      "Strict-Transport-Security": ""
+      "Referrer-Policy": "no-referrer"
+      "X-Content-Type-Options": "nosniff"
+      "X-Download-Options": "noopen"
+      "X-Frame-Options": "SAMEORIGIN"
+      "X-Permitted-Cross-Domain-Policies": "none"
+      "X-Robots-Tag": "noindex, nofollow"
+      "X-XSS-Protection": "1; mode=block"
+
     custom:
     # custom: |-
     #     worker_processes  1;..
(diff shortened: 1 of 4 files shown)


charts/nextcloud/values.yaml
```diff
@@ -328,6 +328,16 @@ nginx:
   config:
     # This generates the default nginx config as per the nextcloud documentation
     default: true
+    header:
```
provokateurin, line 331:
```suggestion
    headers:
```

provokateurin (changes requested):

charts/nextcloud/values.yaml
```diff
@@ -328,6 +328,16 @@ nginx:
...
+      # -- HSTS settings
+      # WARNING: Only add the preload option once you read about
+      # the consequences in https://hstspreload.org/. This option
+      # will add the domain to a hardcoded list that is shipped
+      # in all major browsers and getting removed from this list
+      # could take several months.
+      # example: hsts: "max-age=15768000; includeSubDomains; preload;"
+      hsts:
```
jessebot, line 339:
```suggestion
      hsts: ""
```
Just so we can set a default of empty string, which will still evaluate as null and be ignored by your with in charts/nextcloud/files/nginx.config.tpl, but makes it a tag more clear about typing for those taking a look at values.yaml directly.

jessebot (changes requested):
Can you please also update the README.md?

charts/nextcloud/values.yaml
```diff
@@ -328,6 +328,23 @@ nginx:
...
     # This generates the default nginx config as per the nextcloud documentation
     default: true
+    headers:
+      # -- HSTS settings
+      # WARNING: Only add the preload option once you read about
+      # the consequences in https://hstspreload.org/. This option
+      # will add the domain to a hardcoded list that is shipped
+      # in all major browsers and getting removed from this list
+      # could take several months.
+      # example: hsts: "max-age=15768000; includeSubDomains; preload;"
```
jessebot, line 338: is this still relevant? Should we still include something like this for easy understanding?:

```yaml
config:
  headers:
    hsts: ""
```

jessebot: no, no, it's totally fine to keep the warning. I'm just confused as to how it works. Should we still have something like this:

```yaml
config:
  headers:
    hsts: ""
```

This PR is called "add option to easy set HSTS" and so I just wanted to see an active example in the actual values.yaml and README to help others. Does that make sense? I want to mention that I like this change but just want to be more clear,

charts/nextcloud/values.yaml
```diff
@@ -331,6 +331,23 @@ nginx:
...
```
jessebot, line 341:
```suggestion
      # example: hsts: "max-age=15768000; includeSubDomains; preload;"
      hsts: ""
```

jessebot: thank you that's more clear!

PR #633, opened 2024-09-21
chore(nextcloud): switch to helm-docs

### LINKED ISSUE #547: Discussion: state of helm-docs for generating sectioned documentation
Per my comment to @wrenix [here](https://github.com/nextcloud/helm/pull/529#discussion_r1545202786_):

If we used [helm-docs](https://github.com/norwoodj/helm-docs), that _would_ be cool. It looks like they recently added the ability to do sections (which was largely what stopped us being interested in this before)!
https://github.com/norwoodj/helm-docs/tree/master/example-charts/sections

We'd need to prefix each section with the name of the section like:
```yaml
# @section -- database
postgresql:
# -- enable postgresql sub chart
  enabled: false
```

See their example `values.yaml` here:
https://github.com/norwoodj/helm-docs/blob/master/example-charts/sections/values.yaml

But then you get a specific section you can render in the `README.md` like their example [here](https://github.com/norwoodj/helm-docs/blob/master/example-charts/sections/README.md):

```markdown
## Values

### database

| Key | Type | Default | Description |
|-----|------|---------|-------------|
| postgresql.enabled | bool | `"false"` |  |
```

## Things to consider
- We would also need a [`README.gotmpl`](https://github.com/norwoodj/helm-docs/blob/master/example-charts/full-template/README.md.gotmpl)
  - we would need to import the bulk of the tips and help text in our current readme into the readme template, or it would get deleted anytime someone runs helm-docs
- It would be nice to use a github action to automatically update the docs from the readme. I use [gabe565/setup-helm-docs-action](https://github.com/gabe565/setup-helm-docs-action) for this on my personal repos. It requires write access to at least contents.
- There is a [pre-commit hook for helm-docs](https://github.com/norwoodj/helm-docs/tree/master?tab=readme-ov-file#pre-commit-hook) so we could include that in a `.pre-commit-config.yml` file.

# open question
Can you have comments and additional data for specific sections? For instance, right now, we have a database section of our readme that gives further details on the `externalDatabase` parameter. Is there a way to call the sections in the `README.gotmpl`?

wrenix:
## Description of the change

use helm-docs for documentation

## Benefits

No duplicated documentation between README and values.yaml

## Possible drawbacks

Not yet so good structed as before, TODOs are:

- [x] give every value in `values.yaml` a usefull documentation (split to #693 )
- [x] group to sections
- [ ] add extra documentations on the README helm-docs template
- [ ] add ci-job to check if README is uptodate

## Applicable issues

fixes #547

## Additional information

Readme [rendered](https://github.com/wrenix/nextcloud-helm/blob/fix/helm-docs/charts/nextcloud/README.md)

## Checklist 

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -1,841 +1,395 @@
-# Nextcloud Helm Chart
-
-[Nextcloud](https://nextcloud.com/) is a file sharing server that puts the control and security of your own data back into your hands.
-
-## TL;DR;
-
-```console
-helm repo add nextcloud https://nextcloud.github.io/helm/
-helm install my-release nextcloud/nextcloud
-```
-
-## Quick Links
-
-* [Introduction](#introduction)
-* [Prerequisites](#prerequisites)
-* [Installing the Chart](#installing-the-chart)
-* [Uninstalling the Chart](#uninstalling-the-chart)
-* [Configuration](#configuration)
-    * [Ingress](#ingress)
-        * [Ingress Sticky-Sessions](#ingress-sticky-sessions)
-            * [NGINX Ingress-Controller](#nginx-ingress-controller)
-            * [Traefik Ingress-Controller](#traefik-ingress-controller)
-            * [HAProxy Ingress-Controller (Community-Version)](#haproxy-ingress-controller-community-version)
-    * [Database Configurations](#database-configurations)
-    * [Object Storage as Primary Storage Configuration](#object-storage-as-primary-storage-configuration)
-    * [Persistence Configurations](#persistence-configurations)
-    * [Metrics Configurations](#metrics-configurations)
-    * [Headers set on NGINX](#headers-set-on-nginx)
-    * [Probes Configurations](#probes-configurations)
-    * [Collabora Configuration](#collabora-configuration)
-    * [Imaginary](#imaginary)
-* [Cron jobs](#cron-jobs)
-* [Using the nextcloud docker image auto-configuration via env vars](#using-the-nextcloud-docker-image-auto-configuration-via-env-vars)
-* [Multiple config.php file](#multiple-configphp-file)
-* [Using nginx](#using-nginx)
-    * [Service discovery with nginx and ingress](#service-discovery-with-nginx-and-ingress)
-* [Preserving Source IP](#preserving-source-ip)
-* [Hugepages](#hugepages)
-* [HPA (Clustering)](#hpa-clustering)
-* [Adjusting PHP ini values](#adjusting-php-ini-values)
-* [Running `occ` commands](#running-occ-commands)
-    * [Putting Nextcloud into maintanence mode](#putting-nextcloud-into-maintanence-mode)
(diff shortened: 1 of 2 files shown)


charts/nextcloud/Chart.yaml
```diff
@@ -21,6 +13,17 @@ maintainers:
     email: christian.ingenhaag@googlemail.com
   - name: billimek
     email: jeff@billimek.com
+  - name: WrenIX
+    url: https://wrenix.eu
```
jessebot, line 17: oh oh oh! I wanna be a maintainer too! :D
```suggestion
  - name: WrenIX
    url: https://wrenix.eu
  - name: jessebot
    url: https://jessebot.work
```

charts/nextcloud/Chart.yaml
```diff
@@ -21,6 +13,17 @@ maintainers:
...
+keywords:
+  - nextcloud
+  - storage
+  - http
+  - web
+  - php
+version: 6.0.2
+# renovate: image=docker.io/library/nextcloud
```
jessebot, line 25: 💙

charts/nextcloud/README.md
```diff
@@ -1,698 +1,274 @@
...
-
-### Database Configurations
-By default, nextcloud will use a SQLite database. This is not recommended for production, but is enabled by default for testing purposes. When you are done testing, please set `internalDatabase.enabled` to `false`, and configure the `externalDatabase` parameters below.
-
-For convenience, we packages the following Bitnami charts for databases (feel free to choose _one_ below):
-- [Bitnami MariaDB chart](https://github.com/bitnami/charts/tree/main/bitnami/mariadb)
-- [Bitnami PostgreSQL chart](https://github.com/bitnami/charts/tree/main/bitnami/postgresql)
-
-If you choose to use one of the prepackaged Bitnami helm charts, you must configure both the `externalDatabase` parameters, and the parameters for the chart you choose. For instance, if you choose to use the Bitnami PostgreSQL chart that we've prepackaged, you need to also configure all the parameters for `postgresql`. You do not need to use the Bitnami helm charts. If you want to use an already configured database that you have externally, just set `internalDatabase.enabled` to `false`, and configure the `externalDatabase` parameters below.
```
jessebot, line 214: so right now, I'd worry about losing stuff like this specifically, unless we want to truly separate our docs into it's own wiki or perhaps a doc site? 🤔

provokateurin: Agreed, we lose a lot of the current docs from the README

jessebot: I'm open to using the built in github wiki feature, or playing with the helm-docs templates to make sure we keep these docs, as they help out the users a lot when searching this repo.

provokateurin: I would try to avoid splitting the docs across multiple places (e.g. README and wiki)

charts/nextcloud/Chart.yaml
jessebot: GOOD, mine are like that too! :D

PR #639, opened 2024-09-25
feat(nginx): add ipFamilies option for IPv4 and IPv6

### LINKED ISSUE #636: Nginx container will not listen on IPv6
## Describe your Issue
I am trying to run the chart on a cluster with IPv6 enabled, but the nextcloud pod never completes because of failed startup probe for the nextcloud-nginx container.

The generated default configuration when nginx is enabled only sets the port to 80 which defaults nginx to listen only on IPv4

### Logs and Errors

```
Startup probe failed: Get "http://[2606:re:dac:ted::1:1697]:80/status.php": dial tcp [2606:re:dac:ted::1:1697]:80: connect: connection refused
```
## Describe your Environment

- Kubernetes distribution: Talos v1.7.6
- FluxCD v2.3.0
- Helm Chart Version: 6.0.2

- `values.yaml`:

```yaml
# paste your values.yaml (anonymize any sensitive data)
```

## Additional context, if any

I just recently switched my cluster from IPv4 to dual-stack. I have had the same issue with other deployment and usually changing the listen address to [::]:80 when using nginx resolves the issue. 

Unfortunately, the way the chart generates the default configuration overrides the `/docker-entrypoint.d/10-listen-on-ipv6-by-default.sh` of the nginx:alpine container.

I might submit a PR to the chart to try and fix this by adding a new value, something like `nginx.listenAddress` that could be used in `charts/nextcloud/files/nginx.config.tpl` with the containerPort value.

mdallaire:
## Description of the change

Add an option to configure the listening IP address for the nginx container.

## Benefits

This will allow setting the IP to a value like "[::]" for IPv6 deployments. The current generated configs force the address to 0.0.0.0

## Possible drawbacks

None that I can think of. The default value is 0.0.0.0 which is equivalent to the configuration generated by the chart at the moment.

## Applicable issues

- fixes #636

## Checklist 

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -161,6 +161,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.containerPort`                                       | Customize container port e.g. when not running as root                                              | `IfNotPresent`             |
+| `nginx.ipFamilies`                                          | Customize container to listen on IPv4, IPv6 or both                                                 | `["IPv4"]`                 |
 | `nginx.config.default`                                      | Whether to use nextcloud's recommended nginx config                                                 | `true`                     |
 | `nginx.config.custom`                                       | Specify a custom config for nginx                                                                   | `{}`                       |
 | `nginx.resources`                                           | nginx resources                                                                                     | `{}`                       |
--- a/charts/nextcloud/files/nginx.config.tpl
+++ b/charts/nextcloud/files/nginx.config.tpl
@@ -3,7 +3,16 @@ upstream php-handler {
 }
 
 server {
+    {{- if and (has "IPv4" .Values.nginx.ipFamilies) (has "IPv6" .Values.nginx.ipFamilies) }}
+    # Both IPv4 and IPv6 are enabled
+    listen [::]:{{ .Values.nginx.containerPort }} ipv6only=off;
+    {{- else if and (not (has "IPv4" .Values.nginx.ipFamilies)) (has "IPv6" .Values.nginx.ipFamilies) }}
+    # Only IPv6 is enabled
+    listen [::]:{{ .Values.nginx.containerPort }};
+    {{- else }}
+    # Default, IPv4 only
     listen {{ .Values.nginx.containerPort }};
+    {{- end }}
 
     # HSTS settings
     # WARNING: Only add the preload option once you read about
@@ -48,7 +57,7 @@ server {
     include mime.types;
     types {
         text/javascript js mjs;
-    }        
+    }
 
     # Path to the root of your installation
     root /var/www/html;
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -324,7 +324,10 @@ nginx:
     pullPolicy: IfNotPresent
 
   containerPort: 80
-
+  # This configures nginx to listen on either IPv4, IPv6 or both
+  ipFamilies:
+    - IPv4
+    # - IPv6
   config:
     # This generates the default nginx config as per the nextcloud documentation
     default: true
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 6.0.3
+version: 6.1.0
 appVersion: 30.0.0
 description: A file sharing server that puts the control and security of your own data back into your hands.
 keywords:


provokateurin:
Array sounds like a good idea, if we add it before merging this it also doesn't result in a breaking change.
As for always listening on both, I don't know if that is a good idea. If ipv6 is disabled in the kernel the container will always fail.
So array with both ipv4 and ipv6 by default would make sense to me, then it can still be reduced to ipv4 only if necessary.

provokateurin (changes requested):
Blocking for now

mdallaire:
That would be achieved with the `ipv6only` nginx option. I can add it to the chart.

mdallaire:
It could also be a single option similar to the `ipFamilies` for services. This array with possible values of `IPv4` and `IPv6` or both would decide how both listening address and `ipv6only` is built in the configuration template.

Let me know how you prefer this be done and I will adjust the PR.

charts/nextcloud/README.md
```diff
@@ -161,6 +161,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.containerPort`                                       | Customize container port e.g. when not running as root                                              | `IfNotPresent`             |
+| `nginx.ipFamily`                                       | Customize container to listen on IPv4, IPv6 or both                                                      | `IfNotPresent`             |
```
provokateurin, line 164:
```suggestion
| `nginx.ipFamily`                                           | Customize container to listen on IPv4, IPv6 or both                                                  | `["IPv4"]`             |
```

charts/nextcloud/values.yaml
```diff
@@ -324,7 +324,10 @@ nginx:
     pullPolicy: IfNotPresent
 
   containerPort: 80
-
+  # This configures nginx to listen on either IPv4, IPv6 or both
+  ipFamily:
```
provokateurin, line 328: Shouldn't this be `ipFamilies` since it takes an array which is plural?

charts/nextcloud/files/nginx.config.tpl
```diff
@@ -3,7 +3,17 @@ upstream php-handler {
 }
 
 server {
+    {{- if and (not (has "IPv4" .Values.nginx.ipFamily)) (has "IPv6" .Values.nginx.ipFamily) }}
+    # Only IPv6 is enabled
+    listen [::]:{{ .Values.nginx.containerPort }};
+    {{- else if and (has "IPv4" .Values.nginx.ipFamily) (has "IPv6" .Values.nginx.ipFamily) }}
+    # Both IPv4 and IPv6 are enabled
+    listen [::]:{{ .Values.nginx.containerPort }};
+    ipv6only = off;
```
provokateurin, line 12: I'd prefer the order of these to be the other way around, such that 4+6 use-case is first, then 6 only, and 4 only as fallback.

provokateurin (changes requested):

charts/nextcloud/README.md
```diff
@@ -161,6 +161,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.containerPort`                                       | Customize container port e.g. when not running as root                                              | `IfNotPresent`             |
+| `nginx.ipFamilies`                                           | Customize container to listen on IPv4, IPv6 or both                                                  | `["IPv4"]`             |
```
provokateurin, line 164:
```suggestion
| `nginx.ipFamilies`                                          | Customize container to listen on IPv4, IPv6 or both                                                  | `["IPv4"]`           |
```
Still off :see_no_evil:

provokateurin (changes requested):

charts/nextcloud/README.md
mdallaire: Argh, ok should be good now!

charts/nextcloud/README.md
```diff
@@ -161,6 +161,7 @@ The following table lists the configurable parameters of the nextcloud chart and
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.image.pullPolicy`                                    | nginx Image pull policy                                                                             | `IfNotPresent`             |
 | `nginx.containerPort`                                       | Customize container port e.g. when not running as root                                              | `IfNotPresent`             |
+| `nginx.ipFamilies`                                          | Customize container to listen on IPv4, IPv6 or both                                                  | `["IPv4"]`           |
 | `nginx.config.default`                                      | Whether to use nextcloud's recommended nginx config                                                 | `true`                     |
```
provokateurin, line 165: Indentation is still wrong :/

charts/nextcloud/files/nginx.config.tpl
```diff
@@ -3,7 +3,17 @@ upstream php-handler {
...
+    {{- else if and (not (has "IPv4" .Values.nginx.ipFamilies)) (has "IPv6" .Values.nginx.ipFamilies) }}
+    # Only IPv6 is enabled
+    listen [::]:{{ .Values.nginx.containerPort }};
+    {{- else }}
+    # Default, IPv4 only
     listen {{ .Values.nginx.containerPort }};
+    {{- end }}
```
provokateurin, line 16:
```suggestion
    {{- if has "IPv6" .Values.nginx.ipFamilies }}
    listen [::]:{{ .Values.nginx.containerPort }};
    {{- if has "IPv4" .Values.nginx.ipFamilies }}
    ipv6only = off;
    {{- end }}
    {{- else }}
    # Default, IPv4 only
    listen {{ .Values.nginx.containerPort }};
    {{- end }}
```
I think this logic is a bit simpler

mdallaire: Absolutely!

provokateurin (approved):
All good from my side, but haven't tested it. @jessebot @wrenix can one of you also review?

provokateurin:
(A squash of all commits would be nice to have)

mdallaire:
> (A squash of all commits would be nice to have)

Done!

charts/nextcloud/files/nginx.config.tpl
mdallaire: I will need to revert this, the `ipv6only=off` setting needs to be on the same line as the listen.

provokateurin:
Agreed, we can easily fix this if there is a problem.

mdallaire:
> I am just thinking of dualstack, is not everywhere automatically on if `listen [::]:80;`.

When `listen [::]:80;` is set, nginx will only listen to IPv6 (the default value for the `ipv6only` option is `on` [ref](https://nginx.org/en/docs/http/ngx_http_core_module.html#listen)). This is why we have to set `listen [::]:80 ipv6only=off;` when both IPv4 and IPv6 are present in the array.

Issue #612, opened 2024-07-29
New S3 Configuration breaks prior S3 configuration and is missing uploadPartSize


acidicX:
## Describe your Issue
I've used the S3 integration before "manually" via `configs.s3.config.php`, now the helm chart upgrade (v5.4.0) fails because it's already defined

### Logs and Errors
> Helm upgrade failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:
 line 168: mapping key "s3.config.php" already defined at line 23

## Describe your Environment

- Kubernetes distribution: k8s

- Helm Version (or App that manages helm): 

- Helm Chart Version: 5.4.0

- `values.yaml`:

```yaml
configs:
        s3.config.php: |-
          <?php
          $CONFIG = array (
            'objectstore' => array(
              'class' => '\\OC\\Files\\ObjectStore\\S3',
              'arguments' => array(
                'hostname'         => 's3.eu-central-2.wasabisys.com',
                'bucket'           => '...',
                'autocreate'       => false,
                'key'              => '...',
                'secret'           => '...',
                'region'           => 'eu-central-2',
                'use_ssl'          => true,
                'use_path_style'   => false,
                'uploadPartSize'   => '2097152000'
              )
            )
          );
```

## Additional context, if any

1. It would be nice if such a change would be marked as a breaking change next time
2. The chart should respect the manually set configuration
3. The new options are missing `uploadPartSize`, which is required to be set there for some reason on S3 storage, so I can't even switch to the new configuration style

provokateurin:
Sorry for this breaking change :/ I think your config should continue to work fine by just renaming the config file.
Could you open a PR for the `uploadPartSize` so that it is present and allows you to switch to the new configs?

acidicX:
@provokateurin thanks for the quick response, I can try to open a PR but I'm no expert in helm charts and I can't test the change locally at the moment.

jessebot:
Sorry for the rapid updates, but if #614 gets merged, it will close this Issue automatically. That doesn't mean we can't still help you if other things are still broken here though, so please just let us know if that's the case and we can reopen this issue for you if it gets closed 🙏

acidicX:
@jessebot thanks for the fast reaction, I can confirm that the workaround (`nextcloud.defaultConfigs.s3.config.php=false`) works for me :) I've been able to upgrade and everything seems to work fine.

MaksonHerson:
Hello @jessebot , @acidicX !
[...]

PR #643, opened 2024-10-06
fix(nextcloud): truly not require SSE-C

jstewart612:
## Description of the change

Make SSE-C truly not be required just as the upstream project does not require SSE-C.

## Benefits

Make SSE-C truly not be required just as the upstream project does not require SSE-C.

## Possible drawbacks

None: only makes Helm chart more compliant with available upstream options.

## Applicable issues

- fixes part of #612 

## Checklist 


charts/nextcloud/templates/_helpers.tpl
```diff
@@ -307,7 +307,8 @@ S3 as primary object store env vars
     secretKeyRef:
       name: {{ .Values.nextcloud.objectStore.s3.existingSecret }}
       key: {{ .Values.nextcloud.objectStore.s3.secretKeys.sse_c_key }}
-{{- else }}
+{{- end }}
+{{- if and (gt (len .Values.nextcloud.objectStore.s3.sse_c_key) 0) (lt (len .Values.nextcloud.objectStore.s3.secretKeys.sse_c_key) 1) }}
```
jessebot, line 311:
```suggestion
{{- if .Values.nextcloud.objectStore.s3.sse_c_key }}
```

We don't actually need to calculate the length, as if this is not an empty string, it will trigger. My above suggestion should work fine.

jessebot (changes requested):
Looks good, can be uncomplicated just a bit though :)

wrenix (changes requested):
After the change, you should rebase and fix the version dump.

Issue #577, opened 2024-05-31
Feature: Use CronJob resources instead of a container with crond


pschichtel:
## Description of the change

I would like to use k8s' CronJob resources to run nextcloud's cronjob tasks instead of having them executed by the cron container.

## Benefits

* easier to add new cronjobs (e.g. for certain apps like image preview pre generation or duplicate detection)
* more/easier insight into failed cronjobs through cluster monitoring
* separate resource limits when multiple cronjobs are configured

## Possible drawbacks

I'm not entirely clear on how this might interact with nextcloud during upgrades, but I don't think it changes much compared to the status quo.

provokateurin:
Some long time ago this was how the crons were implemented, but it had quite some problems which I don't remember anymore. Please search in the issues and PRs of this repo to dig up the history of this. If the problems can be avoided now it definitely makes sense to use the k8s cron instead.

pschichtel:
I've found quite a few issues related to cronjobs, but I haven't found any that would suggest an issue with running them as CronJob resources. The cronjob resource used curl to invoke webcron, I can imagine how this could cause various issues like network timeouts, my intent would be to use the same nextcloud/docker image and run the same command the current cron container uses as well.

History of the cronjob.yaml: https://github.com/nextcloud/helm/commits/main/charts/nextcloud/templates/cronjob.yaml

provokateurin:
I guess that makes sense, just execute the cron.php script which can not timeout instead of using curl which can timeout. I wonder why nobody thought about this :sweat_smile:

pschichtel:
If I were to submit a PR to implement this: Should I keep the current cron mechanism around (e.g. introduce an option to select sidecar vs cronjob)?

provokateurin:
I don't think that would be necessary as there is no real breaking change, as long as you use the same container image and mount all the same volumes etc. You're effectively switching https://github.com/nextcloud/docker/blob/13f51c4f70676edb464ce5ea03631403b54623df/29/fpm/Dockerfile#L18 against the k8s Cronjob resource.
@jessebot what do you think?

jessebot:
Yeah, I like the idea of this being a k8s cronjob 👍 I don't see a big a deal with removing the container either.

JuniorJPDJ:
This also allows proper scaling of nextcloud pod, as scaling cron container is a bad idea.

provokateurin:
[...]

PR #740, opened 2025-06-25
feat: add option to run cronjob as CronJob instead of sidecar

Manuelraa:
## Description of the change

Update from PR evolution:
BREAKING CHANGE: `cronjob.command` was renamed to `cronjob.sidecar.command` to avoid confusion with the cronjob command. Please update your `values.yaml` accordingly.

This PR picks up on https://github.com/nextcloud/helm/pull/703
Basing it on newest `main`.
And use existing newest `_helpers` etc.
No unrelated changes.

Additionally it keeps the old `crond` for people where it works fine and they don't want to change.
I'm also fine removing `crond` sidecar. But figured this way the potential impact of the change to existing setups is limited.

I'm willing to make adjustments based on maintainer and user feedback.
README diff looks scary large. But I just added white spaces for formatting.

### Testing
1. `helm install nextcloud .` from `main`
2. Validate nextcloud runs and works
3. Checkout PR branch. `helm upgrade nextcloud . --values /tmp/values.yaml`
4. Confirm nextcloud still works. Validate that cronjob is executed and last reported in nextcloud

image

```
# /tmp/values.yaml
image:
  repository: nextcloud
  tag: 31.0.6-fpm
  # tag: 31.0.5-apache

persistence:
  enabled: true

nginx:
  enabled: true
  image:
    repository: nginxinc/nginx-unprivileged
    tag: alpine
  securityContext:
    runAsUser: 101
    runAsGroup: 33
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL

nextcloud:
  podSecurityContext:
    runAsNonRoot: true
    fsGroup: 33
    fsGroupChangePolicy: "Always"
    seccompProfile:
      type: RuntimeDefault
  securityContext:
    runAsUser: 33
    runAsGroup: 33
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL
  host: mydomain.com
  trustedDomains:
    - localhost
    - nextcloud.nextcloud.svc.cluster.local
    - mydomain.com
  configs:
    basics.config.php: |-
      <?php
      $CONFIG = array (
        'check_data_directory_permissions' => false,
      );

cronjob:
  enabled: true
  useCronJobInsteadOfCrond: true
  securityContext:
    runAsUser: 33
    runAsGroup: 33
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL

```

## Benefits

- Enabled running cron in non-root environments (Fixes `crond: can't set groups: Operation not permitted` with crond running as non-root)

## Possible drawbacks

- Can't think of any

## Applicable issues

- fixes https://github.com/nextcloud/helm/issues/577

## Checklist 


Manuelraa:
Rebased to fix conflicts.

For everyone seeing this: Using this for 3 weeks without issues so far.

@wrenix could you please review this PR?

charts/nextcloud/values.yaml
```diff
@@ -568,7 +566,8 @@ cronjob:
 
   # Allow configuration of lifecycle hooks
   # ref: https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
-  lifecycle: {}
+  lifecycle:
+    {}
```
DreamingRaven, line 570: Could you also return this one to the same standard please, sorry for the hassle.

charts/nextcloud/values.yaml
```diff
@@ -625,7 +654,8 @@ persistence:
     accessMode: ReadWriteOnce
     size: 8Gi
 
-resources: {}
+resources:
+  {}
```
DreamingRaven, line 658: Could you return this one to normal as well please.

charts/nextcloud/values.yaml
```diff
@@ -580,7 +579,37 @@ cronjob:
...
+    #             operator: In
+    #             values:
+    #               - app
+    #       topologyKey: kubernetes.io/hostname
+  commandCronJob:
+    - php
+    - -f
+    - /var/www/html/cron.php
+    - --
+    - --verbose
```
DreamingRaven, line 612: I feel like someone is going to get confused between the two commands, however im not sure if there is anything we can do to prevent confusion here.

DreamingRaven (changes requested):
Thanks for the great PR, just a few amendments I would recommend.

I would have reviewed this sooner, but I only just got permissions today!

charts/nextcloud/templates/cronjob.yaml
```diff
@@ -0,0 +1,136 @@
...
+        {{- end }}
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          containers:
+            - name: {{ .Chart.Name }}-cron
+              image: {{ include "nextcloud.image" . }}
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              command:
```
varet80, line 35: I suggest to change the command to:
```
runuser -u www-data -- /usr/local/bin/php -d php.memory_limit=2G cron.php
```
1. set memory limit and add option to tune it
2. use runuser to run it as www-data, otherwise files will change to root ownership in some cases.

charts/nextcloud/templates/cronjob.yaml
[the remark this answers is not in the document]
Manuelraa: > the concurrency forbid seems fine, it can get locked up, but I am not 100% certain on the current behaviour so I might have to check that, but its probably better to let it complete and potentially lock up than always kill the old without it completing / getting interrupted all the time."Forbid":
    forbids concurrent runs, skipping next run if previous run hasn't finished

Spec says:
```
"Forbid": forbids concurrent runs, skipping next run if previous run hasn't finished yet;
```

We could add `activeDeadlineSeconds`. Not sure if the Nextcloud cron is "transaction safe" and can just be killed.
A normal crond would not terminate the running cmd either?

What would happen here is no new job would be created until the previous is done (e.g. long/slow task)

Long story short: Happy to set it to `Allow` parallel runs and/or set `activeDeadlineSeconds`.

charts/nextcloud/values.yaml
Manuelraa: I get the point. But I feel like `external` could be confusing as well. As per "outside of this chart" etc.
I can't come up with better name than `cronjob.cronjob` maybe `cronjob.kubernetesCronJob`. But that feels weird/long as well.

Implemented the suggestion.

Manuelraa:
Same thing. Implemented suggestion for `crobjob.type` and deployed to my infra.
Also generated with default values `helm template .` before and after PR and no changes visible besides version change.

charts/nextcloud/values.yaml
DreamingRaven: Hey apologies, I will have to get back to you on testing this one, I have hit rate limits and I cannot properly template and test the helm chart as I cannot pull half of the chart's dependencies.

Partly due to this change just announced and trying to prepare for it: https://github.com/bitnami/containers/issues/83267 which also affects the dependencies of this chart. :/ 6 hour wait for this rate limit is killer!

charts/nextcloud/templates/deployment.yaml
```diff
@@ -240,13 +240,14 @@ spec:
...
+        {{- with .Values.cronjob.sidecar }}
+        - name: {{ $.Chart.Name }}-cron
           image: {{ include "nextcloud.image" . }}
-          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          imagePullPolicy: {{ $.Values.image.pullPolicy }}
           command:
-            {{- toYaml .Values.cronjob.command | nindent 12 }}
-          {{- with .Values.cronjob.lifecycle }}
+            {{- toYaml .command | nindent 12 }}
+          {{- with .lifecycle }}
```
DreamingRaven, line 250: @Manuelraa apologies for the delay. I have finally managed to get back on this after sorting out the Bitnami mess.

When the default cronjob sidecar is enabled:
```diff
 cronjob:
-  enabled: false
+  enabled: true
```
The chart fails templating with the following:
```bash
> helm template charts/nextcloud
Error: template: nextcloud/templates/deployment.yaml:246:20: executing "nextcloud/templates/deployment.yaml" at <include "nextcloud.image" .>: error calling include: template: nextcloud/templates/_helpers.tpl:45:14: executing "nextcloud.image" at <.Values.image.tag>: nil pointer evaluating interface {}.image

Use --debug flag to render out invalid YAML
```

Which looks like is a result of changing the context using `with` which includes the `{{ include "nextcloud.image" . }}` You will probably want to change the context to be what was outside of the with statement.

DreamingRaven (changes requested):
Great PR, just that one sentence and I will approve it. (and merge conflicts)

Manuelraa:
> Great PR, just that one sentence and I will approve it. (and merge conflicts)

@DreamingRaven Review comment implemented and rebased. Thank you very much.

DreamingRaven (approved):
LGTM.
Further tested locally with different variations to generate expected output and it passes kubeconform.

PR #780, opened 2025-10-02
chore: update container to latest image (32.0.0)

varet80:
Update image to 32.0.0


varet80:
@wrenix I agree had the same thought after I submit it, as this is major change.
I would say Minor is better, as major I would keep it only for changes like: changing Database backend support

PR #789, opened 2025-10-17
fix(nextcloud): Add support for `priorityClassName`

### LINKED ISSUE #178: PriorityClassName missing from deployment
I would like to define a [priorityClassName](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#pod-priority) for the nextcloud deployment and the cronjob. Pod priority ensures that the pod is not being evicted by other pods.

KarlJorgensen:
## Description of the change

The chart now has one "main" `.Values.priorityClassName`, which is the default for everything deployed in the main chart (i.e. the default does not extend to `mariadb`, `redis`, `collabora-online` etc).

The main priority class can be overridden for the cronjob and imaginary

## Applicable issues

- fixes #178

## Checklist 

### DIFF
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -785,8 +785,8 @@ kubectl exec $NEXTCLOUD_POD -- su -s /bin/sh www-data -c "php occ myocccomand"
 
 Here are some examples below.
 
-### Putting Nextcloud into maintanence mode
-Some admin actions require you to put your Nextcloud instance into [maintanence mode](https://docs.nextcloud.com/server/latest/admin_manual/maintenance/backup.html#maintenance-mode) (e.g. backups):
+### Putting Nextcloud into maintenance mode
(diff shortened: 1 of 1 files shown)


charts/nextcloud/values.yaml
```diff
@@ -666,6 +670,11 @@ cronjob:
     ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
     ##
     resources: {}
+
+    # priorityClassName: priority class for the cron job. Overrides
```
wrenix: Please use helm-docs annotation
```suggestion
    # -- priorityClassName: priority class for the cron job. Overrides
```

charts/nextcloud/values.yaml
```diff
@@ -744,6 +753,12 @@ resources: {}
 #    cpu: 100m
 #    memory: 128Mi
 
+# priorityClassName: Priority class for pods. This is the _default_
```
wrenix: Helm-docs format for all comments

KarlJorgensen: sorry - i don't follow your comment?

PR #792, opened 2025-10-21
Add namespace to templates

### LINKED ISSUE #791: Missing namespace on Helm manifests cause issues with Kustomize
Many Helm resources do not specify the namespace, which can cause issues with Kustomize in some deployments.  I have to add it at the "top" level, which automatically applies the desired namespace to all Kustomize resources. Unfortunately, I have a few resources that have to be applied to a different namespace, which this breaks.

I have highlighted the issue in my Nextcloud deployment manifests: : https://github.com/kenlasko/k8s/blob/main/manifests/apps/nextcloud/base/kustomization.yaml

I should be able to define the namespace at the Helm chart level and it should apply the namespace to all namespaced resources. 

I recommend that:
```
namespace: {{ .Release.namespace }}
```
be added to all namespaced resources, such as:
- deployment
- secret
- service
- serviceAccount

I'd offer to do this myself, but I'm not a developer and not familiar with the process to do such things.

EDIT: Think I managed to submit a pull request to fix this for all objects. Crossing fingers!

kenlasko:
## Description of the change
Add metadata.namespace: {{ .Release.Namespace }} to all namespaced templates

## Benefits
When using Kustomize to render the Helm chart, the namespace defined in the Helm chart doesn't get applied to all Helm-generated resources unless you explicitly tell Kustomize to force-add it to everything. This may not be desired in some situations where manifests might be included in the Kustomization bound to other namespaces. 

## Possible drawbacks
No known drawbacks.

## Applicable issues
#791 
- fixes #791 

## Additional information
[See here](https://github.com/kenlasko/k8s/blob/main/manifests/apps/nextcloud/base/kustomization.yaml) for an example of what I mean.

## Checklist 


kenlasko:
Works as expected. Email came through fine with the attachment intact. 

One other change is to move the Save Destinations button to be directly under the S3 settings and add a Test button for verification. Changing the other settings should not require pressing a Save button

PR #806, opened 2025-11-09
fix(values): set registry

volker-raschek:
The following patch adds the `registry` attribute. It also sets the `registry` attribute for all helm chart dependencies.

This is necessary to protect CRIO users from a breaking change in v1.34. Further information can be found in the following [pr](https://github.com/cri-o/cri-o/pull/9401) of cri-o.

  Volker


wrenix (commented):
Lets also add also a global image option to overwrite it for all the images
```yaml
global:
  image:
   # -- if set it will overwrite all registry entries
    registry:
```

PR #816, opened 2025-11-29
Add dataVolumeMount option

Javex:
## Description of the change

By default, when persistence is enabled, the volume "nextcloud-main" is used to store Nextcloud data. The chart offers an option to create an extra PVC using `persistence.nextcloudData`.

However, it's not possible to mount data from an external source. Even though `extraVolumes` and `extraVolumeMounts` are available, the data folder is either sourced from "nextcloud-main" or "nextcloud-data".

By providing an option to skip mounting the "data" subfolder when using the main persistence option, but not the `nextcloudData` option, it is possible to provide a custom volume mount based on a different volume specified in `extraVolumes`.

This allows, for example, to supply an NFS mount point:

```yaml
nextcloud:
  extraVolumes:
    - name: data
      nfs:
        path: /mnt/nextcloud
        server: my-nfs-server.example.com

  extraVolumeMounts:
    - name: data
      mountPath: /var/www/html/data

persistence:
  # Create a PVC to store Nextcloud application data
  enabled: true
  # But don't mount the user data folder, as it is provided above
  dataVolumeMount: false
```

## Benefits

Users with specific storage setup can customise how their data gets mounted. For example, I have an existing Nextcloud installation that I want to migrate to Kubernetes, but the data is already in an NFS server and I just want to mount that.

## Possible drawbacks

It's  possible I missed a configuration combination that doesn't make sense here, there are many ways to combine options in this chart.

## Applicable issues

N/A

## Additional information

I've first tried to use the NFS CSI driver, but it caused quite a few issues. This would have worked without needing to change the helm chart. However, it did not work out due to all problems this caused me.

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 8.6.0
+version: 8.7.0
 # renovate: image=docker.io/library/nextcloud
 appVersion: 32.0.2
 description: A file sharing server that puts the control and security of your own data back into your hands.
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -405,6 +405,7 @@ Nextcloud will *not* delete the PVCs when uninstalling the helm chart.
 | `persistence.existingClaim`               | An Existing PVC name for nextcloud volume            | `nil`           |
 | `persistence.accessMode`                  | PVC Access Mode for nextcloud volume                 | `ReadWriteOnce` |
 | `persistence.size`                        | PVC Storage Request for nextcloud volume             | `8Gi`           |
+| `persistence.dataVolumeMount`             | Toggle automatic mount of data volume                | `true`          |
 | `persistence.nextcloudData.enabled`       | Create a second PVC for the data folder in nextcloud | `false`         |
 | `persistence.nextcloudData.annotations`   | see `persistence.annotations`                        | `{}`            |
 | `persistence.nextcloudData.labels`       | see `persistence.labels`                             | `{}`            |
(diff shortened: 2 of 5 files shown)


BinaryMan32:
It looks like both main and data mounts support `existingClaim`, so while it's just slightly more work you should be able to create a PVC in addition to your PV and use something like:
```
persistence:
  enabled: true
  existingClaim: my-main-pvc
  nextcloudData:
    enabled: true
    existingClaim: my-data-pvc
```
See [templates/deployment.yaml](https://github.com/nextcloud/helm/blob/83e1bbb1c83c098c54eb05ff17f223274a6beac0/charts/nextcloud/templates/deployment.yaml#L352-L375) which in this case should end up generating something similar to:
```yaml
        - name: nextcloud-main
          persistentVolumeClaim:
            claimName: my-main-pvc
        - name: nextcloud-data
          persistentVolumeClaim:
            claimName: my-data-pvc
```
While it means you have to create one additional resource (PVC), in exchange you don't have to worry about configuring the mount path yourself like you would have to with `extraVolumeMounts`.

To create PVC which is bound to an existing PV, see [Reserving a PersistentVolume](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reserving-a-persistentvolume).

Javex:
Thanks for the suggestion wrenix, when I have some time I'll try and refactor this PR like you suggested. I also dislike my approach, I just couldn't think of anything better.

@BinaryMan32 , my NFS volume is **not** a PV. It's a direct NFS mount. I tried going the PVC route first, but:
> I've first tried to use the NFS CSI driver, but it caused quite a few issues. This would have worked without needing to change the helm chart. However, it did not work out due to all problems this caused me.

BinaryMan32:
> my NFS volume is not a PV. It's a direct NFS mount

@Javex apologies, I missed that. If you've already mounted the NFS share outside of kubernetes, then don't you essentially have a `hostPath` which is supported by the chart?

Edit: If you haven't already mounted the NFS share outside the chart, I don't think there is any functional difference in terms of how NFS is used whether you configure directly via `nfs` under `extraVolumes` or in a `PV`. I understand that it's annoying to have to create the additional `PV`/`PVC` resources, but if you are creating the `PV`/`PVC` yourself with `storageClassName: ""` then the NFS CSI driver shouldn't be involved.

PR #851, opened 2026-04-27
Making use of .nginx.containerPort when fpm is enabled

Ejdesgaard:
This allows different nginx and fpm ports to be used, by utilizing the `.nginx.containerPort`, when fpm is enabled.


Ejdesgaard:
Sorry for all the force-pushes, I tried to sort out the commit message...

wrenix:
Why not drop nginx.container port ?
On Service and probes we should change to port-name instatt of port-number.

PR #855, opened 2026-05-04
fix(nextcloud): enable nginx config fastcgi_request_buffering

### LINKED ISSUE #854: Nginx fastcgi_request_buffering should be enabled
## Description of the change

Some WebDAV clients can not upload files when Nginx and php-fpm is used: https://github.com/nextcloud/documentation/issues/9574

To fix it the [Nextcloud admin manual](https://docs.nextcloud.com/server/stable/admin_manual/installation/nginx.html) suggests to set nginx configuration `fastcgi_request_buffering on` (which is also the nginx default behavior), but the helm chart set it to `off`.

## Benefits

When fastcgi_request_buffering is enabled, the entire request body is read from the client before sending the request to a FastCGI server.

Nginx adds a Content-Length header when chunked encoding is used to fix WebDAV client issues
Also it reduces timeouts during large uploads caused by php-fpm and nginx may shield php-fpm from resource exhaustion.

When fastcgi_request_buffering is disabled, the request body is sent to php-fpm immediately as it is received.

## Possible drawbacks

Nginx writes the client requests body to a temporary file when it exceeds client_body_buffer_size (per default 8K or 16K depending on the platform).
Nginx could write temporary files on-disk as large as client_max_body_size:

https://github.com/nextcloud/helm/blob/d4f84d9be0d1774d8c9119ba87f2a9ca4379b2d3/charts/nextcloud/values.yaml#L417

## Additional information
(none)

RisingOpsNinja:
## Description of the change

Fixes https://github.com/nextcloud/helm/issues/854

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 9.0.6
+version: 9.1.0
 # renovate: image=docker.io/library/nextcloud
 appVersion: 33.0.3
 description: A file sharing server that puts the control and security of your own data back into your hands.
--- a/charts/nextcloud/files/nginx.config.tpl
+++ b/charts/nextcloud/files/nginx.config.tpl
@@ -149,7 +149,7 @@ server {
         fastcgi_pass php-handler;
 
         fastcgi_intercept_errors on;
-        fastcgi_request_buffering off;
+        fastcgi_request_buffering on;
 
         fastcgi_max_temp_file_size 0;
     }


wrenix:
I believe we should change back `client_max_body_size` either, if with this patch works chunk upload is working.

RisingOpsNinja:
> I believe we should change back `client_max_body_size` either, if with this patch works chunk upload is working.

@wrenix I agree, 10GB is an exceptional large default value. Should the change be a separate pull request?

wrenix:
I dont care. If you linke, then there.


After rebase you need to increase the Version in `Chart.yaml` again

wrenix (approved):

PR #864, opened 2026-06-25
fixt(nextcloud): extract db environment

wrenix:
## Description of the change

extract db environment variables for easier reuse

## Benefits
easier reuse e.g. for notify_push later or with `extraManifests`
## Applicable issues

- fixes #

## Checklist 

### DIFF
--- a/charts/nextcloud/templates/_helpers.tpl
+++ b/charts/nextcloud/templates/_helpers.tpl
@@ -64,11 +64,7 @@ Create image name that is used in the deployment
 {{/*
 Create environment variables used to configure the nextcloud container as well as the cron sidecar container.
 */}}
-{{- define "nextcloud.env" -}}
-{{- if .Values.phpClientHttpsFix.enabled }}
-- name: OVERWRITEPROTOCOL
-  value: {{ .Values.phpClientHttpsFix.protocol | quote }}
-{{- end }}
+{{- define "nextcloud.env.database" -}}
 {{- if .Values.internalDatabase.enabled }}
 - name: SQLITE_DATABASE
   value: {{ .Values.internalDatabase.name | quote }}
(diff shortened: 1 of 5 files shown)


charts/nextcloud/templates/_helpers.tpl
```diff
@@ -87,6 +83,8 @@ Create environment variables used to configure the nextcloud container as well a
     secretKeyRef:
       name: {{ .Values.externalDatabase.existingSecret.secretName | default (printf "%s-db" .Release.Name) }}
       key: {{ .Values.externalDatabase.existingSecret.passwordKey }}
+- name: DATABASE_URL
+  value: "mysql://$(MYSQL_USER):$(MYSQL_PASSWORD)@$(MYSQL_HOST)/$(MYSQL_DATABASE)"
```
DreamingRaven, line 87: Does this work? I did not think you could resolve these at runtime like this unless the application itself understood to substitute envs.

DreamingRaven (approved):
LGTM assuming the new substitution works. My validation checks pass with kubeconform, and nothing that I can see that should be breaking.

The README could probably use regenerating since you changed the values.yaml but its so minor in this case I dont think it matters much either way.

charts/nextcloud/templates/_helpers.tpl
wrenix: Dependent environment variables are part of kubernetes:
https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/

This is with not braces `{}` for subsitution, this are brackets `()`.

DreamingRaven: Oooh I see, somehow in my years of writing kubernetes manifests I had not seen this before, a very useful feature that I wish I knew about sooner! Thanks for sharing the link too!

PR #866, opened 2026-07-02
feat(nextcloud): add CalDAV/CardDAV redirects to HTTPRoute

### LINKED ISSUE #860: Feature: HTTPRoute Add caldav and webdav redirect rules
## Description of the change

There is documentation for ingress-nginx to configure caldav and webdav redirects

https://github.com/nextcloud/helm/tree/main/charts/nextcloud#service-discovery-with-nginx-and-ingress

With gateway-api we can define this by-default because no implementation specific config is needed.

It is also not currently possible to define this configuration because backendRefs are always templated on to rules, which is not allowed for redirect filters:

```
 HTTPRoute/nextcloud: HTTPRoute.gateway.networking.k8s.io "nextcloud" is invalid: spec.rules[1]: Invalid value: RequestRedirect filter must not be used together with backendRefs
```

## Benefits

Properly setup caldav and webdav redirects for all gateway api users

## Possible drawbacks

None

somaz94:
## Description of the change

The chart's `HTTPRoute` template (`route.yaml`) always added `backendRefs` to every rule, which made it impossible to define a rule using a `RequestRedirect` filter (Gateway API rejects `RequestRedirect` combined with `backendRefs`). This PR:

- Only renders `backendRefs` for rules that do not use a `RequestRedirect` filter, so redirect rules can now be defined.
- Adds the standard Nextcloud service discovery redirects (`/.well-known/carddav` and `/.well-known/caldav` -> `/remote.php/dav`) as `RequestRedirect` rules, gated behind `httpRoute.wellKnown.enabled` (on by default). With Gateway API these need no implementation-specific config.

## Benefits

CalDAV/CardDAV service discovery works out of the box for Gateway API users, matching the ingress-nginx server-snippet behaviour already documented in the README.

## Possible drawbacks

`httpRoute.wellKnown.enabled` defaults to `true`, so users who already enable `httpRoute` get the two redirect rules added on upgrade. Set `httpRoute.wellKnown.enabled: false` to opt out.

## Applicable issues

- fixes #860

## Additional information

Validated locally with `helm lint` and `helm template` for: default (redirects + catch-all), `wellKnown.enabled=false`, and a user-supplied `RequestRedirect` rule (confirmed no `backendRefs` is rendered).

## Checklist

- [x] I have read the CONTRIBUTING.md doc.
- [x] DCO has been signed off on the commit.
- [x] Chart version bumped in `Chart.yaml` according to semver.

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 9.2.1
+version: 9.2.2
 # renovate: image=docker.io/library/nextcloud
 appVersion: 34.0.1
 description: A file sharing server that puts the control and security of your own data back into your hands.
--- a/charts/nextcloud/README.md
+++ b/charts/nextcloud/README.md
@@ -707,6 +707,9 @@ ingress:
         deny all;
       }
 ```
+
+When exposing Nextcloud through the Gateway API (`httpRoute.enabled: true`), the CalDAV and CardDAV redirects are added automatically as `RequestRedirect` rules (`httpRoute.wellKnown.enabled`, on by default), so no implementation-specific configuration is required. Set `httpRoute.wellKnown.enabled: false` to opt out.
+
 ## Preserving Source IP
 
 - Make sure your loadbalancer preserves source IP, for bare metal, `metalb` does and `klipper-lb` doesn't.
--- a/charts/nextcloud/templates/route.yaml
+++ b/charts/nextcloud/templates/route.yaml
@@ -25,13 +25,37 @@ spec:
     {{- tpl (toYaml .) $ | nindent 4 }}
   {{- end }}
   rules:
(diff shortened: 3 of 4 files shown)


varet80:
@somaz94 how come this is still draft? 
are there more improvements coming in?

somaz94:
@varet80 No more changes coming — it was in draft only while I validated the CI matrix locally and on CI. Everything's green now, so I've marked it ready for review. Thanks for the ping!

varet80 (approved):

varet80:
@somaz94 can you bump patch version of helm?

varet80 (changes requested):
Please bump helmchart version

somaz94:
Bumped the chart patch version to `9.2.2` and rebased on `main` (also integrated the explicit `group`/`kind` on `backendRefs` from the recent Argo CD drift fix). Thanks for the review!

varet80 (approved):

PR #871, opened 2026-07-20
fix(cronjob): change sidecar to cronstable for nonroot compatible (#733)

### LINKED ISSUE #733: nextcloud-cron does not work in OpenShift
## Describe your Issue

In OpenShift kubernetes cluster, with restricted-v2 pod policy, the UID is within a range. The volume data belongs to the first UID of this range, instead of www-data, which will result in a permission error.

### Logs and Errors

```
crond: USER www-data pid 333 cmd php -f /var/www/html/cron.php
crond: can't set groups: Operation not permitted
crond: USER www-data pid 337 cmd php -f /var/www/html/cron.php
crond: can't set groups: Operation not permitted
crond: USER www-data pid 341 cmd php -f /var/www/html/cron.php
crond: can't set groups: Operation not permitted
crond: USER www-data pid 342 cmd php -f /var/www/html/cron.php
```

## Describe your Environment

- Kubernetes distribution: 
OKD 4.15
- Helm Version (or App that manages helm): 

- Helm Chart Version: 
nextcloud helm 6.6.7
- `values.yaml`:

```yaml
cronjob:
  #enabled: false
  enabled: true
  #securityContext: {}
  securityContext:
    runAsUser: null
    runAsGroup: null
    runAsNonRoot: true
```

## Additional context, if any

Inside the container, some info:
```
1004080000@my-nextcloud-6b9695ccfb-spplh:/var/www/html$ id
uid=1004080000(1004080000) gid=0(root) groups=0(root),1004080000

1004080000@my-nextcloud-6b9695ccfb-spplh:/var/www/html$ ls -al /var/www/html/cron.php
-rw-r--r--. 1 1004080000 1004080000 7550 May 27 12:50 /var/www/html/cron.php

1004080000@my-nextcloud-6b9695ccfb-spplh:/var/www/html$ cat /etc/passwd | grep -e www-data -e 1004080000
www-data:x:33:33:www-data:/var/www:/usr/sbin/nologin
1004080000:x:1004080000:0:1004080000 user:/:/sbin/nologin

1004080000@my-nextcloud-6b9695ccfb-spplh:/var/www/html$ cat /var/spool/cron/crontabs/www-data
*/5 * * * * php -f /var/www/html/cron.php
```

antoinetran:
## Description of the change

In shared Kubernetes environment which disable root containers, the cron as sidecar cannot work because cron implementation from alpine requires root. This change will replace to cronstable which is docker compliant and does not need root to work.

## Benefits

Can work in restricted Kubernetes environment:
- with non root policy
- without PV as readwrite many, so impossible to use cronjob as Kubernetes cronjob

## Possible drawbacks

Adds another image, but small footprint (~40MiB for alpine flavor)

## Applicable issues

- fixes #733

## Checklist 

### DIFF
--- a/charts/nextcloud/values.yaml
+++ b/charts/nextcloud/values.yaml
@@ -680,6 +680,16 @@ cronjob:
     ##
     resources: {}
 
+    image:
+      # -- cronstable image registry # can be docker.io and ghcr.io. docker.io has a rate-limit.
+      registry: ghcr.io
+      # -- cronstable image name
+      repository: ptweezy/cronstable
+      # -- cronstable image tag
+      tag: 1.2.24-alpine
+      # -- cronstable image pull policy
+      pullPolicy: IfNotPresent
+
     # Allow configuration of lifecycle hooks
     # ref: https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/
     lifecycle: {}
@@ -693,9 +703,15 @@ cronjob:
     #   runAsNonRoot: true
     #   readOnlyRootFilesystem: true
 
-    # The command the cronjob container executes.
+    schedule: "*/5 * * * *"
+    # The command to run in the cronjob container
+    # Example to increase memory limit: php -d memory_limit=2G ...
     command:
-      - /cron.sh
+      - php
+      - -f
+      - /var/www/html/cron.php
+      - --
+      - --verbose
 
   # Uses a Kubernetes CronJob to execute the Nextcloud cron tasks
   # Note: can run as non-root user. Should run as same user as the Nextcloud pod.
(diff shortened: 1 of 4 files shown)


charts/nextcloud/values.yaml
```diff
@@ -680,6 +680,16 @@ cronjob:
     ##
     resources: {}
 
+    image:
+      # -- cronstable image registry # can be docker.io and ghcr.io. docker.io has a rate-limit.
+      registry: ghcr.io
+      # -- cronstable image name
+      repository: ptweezy/cronstable
```
wrenix, line 687: i am not sure, if we like to have more images from small projects here.

if you need a different compatible you could go to k8s cronjobs.

antoinetran: I understand.

> if you need a different compatible you could go to k8s cronjobs.

I tried it before, it did not work because the way nextcloud implements cronjob requires PVC with ReadWriteMany, which is not possible in any kubernetes cluster.

antoinetran: I also tried to ask busybox cron and vivix cron if they can provide a non-root compatible crond and this is not an easy task. So that's why I tried another image, which is less popular than busybox. At least this one is an official image, compared to supercronic, which also work as non-root but they do not provide official image.

PR #752, opened 2025-07-17
Fix #187: fix redis-session.ini permission denied

antoinetran:
## Description of the change

Following closure of https://github.com/nextcloud/helm/pull/717 because of target branch develop deleted. This PullRequest targets main branch now.

## Applicable issues

- fixes #

## Checklist 

- [x] I have read the [CONTRIBUTING.md](https://github.com/nextcloud/helm/blob/main/CONTRIBUTING.md#pull-requests) doc.
- [x] DCO has been [signed off on the commit](https://docs.github.com/en/github/authenticating-to-github/signing-commits).
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] (optional) Parameters are documented in the README.md

PR #872, opened 2026-07-22
fix: nil pointer when evaluating when cronjob.enabled=true

sapslaj:
## Description of the change

The `if` block checking if Redis is enabled needs to be anchored correctly since we are inside a `with` block (`with .Values.cronjob.sidecar`). Otherwise you get an error:

```
Error: nextcloud/templates/deployment.yaml:274:26
  executing "nextcloud/templates/deployment.yaml" at <.Values.redis.enabled>:
    nil pointer evaluating interface {}.redis
```

## Benefits

Makes the chart renderable with `--set cronjob.enabled=true`.

## Possible drawbacks

I believe this is the only occurrence but I might have missed an `{{- if .Values.redis.enabled }}` that needs to be `{{- if $.Values.redis.enabled }}` instead in some branch. I went through all the places where it was introduced in #752 but might have missed others.

## Applicable issues

- issue introduced in #752

## Checklist 

### DIFF
--- a/charts/nextcloud/Chart.yaml
+++ b/charts/nextcloud/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: nextcloud
-version: 9.2.3
+version: 9.2.4
 # renovate: image=docker.io/library/nextcloud
 appVersion: 34.0.1
 description: A file sharing server that puts the control and security of your own data back into your hands.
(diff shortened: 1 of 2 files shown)


varet80:
LGTM

varet80 (approved):

Issue #814, opened 2025-11-22
Container don't start on RKE2


Ginkan616:
## Describe your Issue
I try to start nexcloud on my RKE2 cluster but after a lot of test with the values i have always the same issue.

### Logs and Errors
 kubectl logs -n cloud  nextcloud-5fbd5d748f-gsnjj --all-containers --tail=200
/entrypoint.sh: 118: cannot create /usr/local/etc/php/conf.d/redis-session.ini: Permission denied
Configuring Redis as session handler
crond: crond (busybox 1.37.0) started, log level 8
crond: can't set groups: Operation not permitted
crond: USER www-data pid   7 cmd php -f /var/www/html/cron.php

## Describe your Environment

- Kubernetes distribution: RKE2

- Helm Version (or App that manages helm): 
version.BuildInfo{Version:"v3.17.3", GitCommit:"e4da49785aa6e6ee2b86efd5dd9e43400318262b", GitTreeState:"clean", GoVersion:"go1.23.7"}

- Helm Chart Version: 
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
nextcloud       cloud           1               2025-11-22 22:02:43.438581025 +0100 CET deployed        nextcloud-8.5.9 32.0.2

- `values.yaml`:

```yaml
# values.yaml — chart: nextcloud/nextcloud

nextcloud:
  host: fqdn
  username: admin
  password: "XXX"
  configs:
    10-overrides.config.php: |-
      <?php
      $CONFIG = array (
        'overwrite.cli.url' => 'https://fqdn',
        'overwriteprotocol' => 'https',
      );

image:
  repository: nextcloud
  tag: 32.0.1-apache

persistence:
  enabled: true
  existingClaim: pvc-cloud

internalDatabase:
  enabled: false
mariadb:
  enabled: false

# Base externe PostgreSQL
externalDatabase:
  enabled: true
  type: postgresql
  host: "my-pgvecto-db-postgresql.database.svc.cluster.local"
  port: 5432
  user: "nextcloud"
  database: "nextcloud"
  existingSecret:
    enabled: true
    secretName: nextcloud-db-secret
    passwordKey: db-password

ingress:
  enabled: true
  className: nginx
  annotations:
    kubernetes.io/tls-acme: "true"
    cert-manager.io/cluster-issuer: "letsencrypt-prod"
  tls:
    - hosts: [fqdn]
      secretName: secret

redis:
  enabled: true
[...]

Issue #187, opened 2022-01-25
/entrypoint.sh: line 192: php/conf.d/redis-session.ini: Permission denied for non-root container


devent:
Need to run a non-root container. I set the `securityContext` as following:

```
securityContext:
  runAsUser: 33
  runAsGroup: 33
```

Get the error:

```
Configuring Redis as session handler
/entrypoint.sh: 56: cannot create /usr/local/etc/php/conf.d/redis-session.ini: Permission denied
```

Of course a non-root can not write to `/usr/local/etc/php/conf.d`

Source https://github.com/nextcloud/docker/blob/master/22/apache/entrypoint.sh#L77

jessebot:
Does this still happen today using the latest version of this helm chart? I know that config may always be owned by root, as per this issue, https://github.com/nextcloud/helm/issues/335, however, I'm not sure that's related. You could try setting this:

```yaml
nextcloud:
  podSecurityContext:
    runAsUser: 33
    runAsGroup: 33
    runAsNonRoot: true
    fsGroup: 33
```

However, that may already get set here:

https://github.com/nextcloud/helm/blob/ff07aacd1bb2199e36544e11661835157cc7a6a3/charts/nextcloud/templates/deployment.yaml#L346-L364

**Note**: Since this issue was introduced, we've modified the available `securityContext` parameters. We have `nextcloud.securityContext`, `nginx.securityContext` (for the containers) and `nextcloud.podSecurityContext` (for the pod).

jilleJr:
@jessebot Still having the issue when trying your YAML snippet. Also can't use the example from the chart's values.yaml as they are invalid values:

https://github.com/nextcloud/helm/blob/0b6bd840f2757e7737566bf388f0e45aa56467d9/charts/nextcloud/values.yaml#L195-L208

Have to revert to last known working version, which for me is  3.5.19 of the chart.

jessebot:
@jilleJr or @devent could you please provide your whole values.yaml after anonymizing any sensitive data?

>  Also can't use the example from the chart's values.yaml as they are invalid values

Could you explain what you mean by invalid and where you're seeing the error? Those are the default values that are applied if you're using the non-alpine containers.

jilleJr:
> @jilleJr or @devent could you please provide your whole values.yaml after anonymizing any sensitive data?
> 
> > Also can't use the example from the chart's values.yaml as they are invalid values

Sure:

<details>
<summary>values.yaml</summary>

```yaml
image:
[...]

PR #879, opened 2026-08-15
fix: Mount the writable emptyDir file into the nextcloud container directly

edvinas31:
## Description of the change

Mount the writable php-confd emptyDir over /usr/local/etc/php/conf.d/redis-session.ini in the main nextcloud container, guarded by .Values.redis.enabled.

When redis.enabled=true the chart already creates everything needed for this:

a php-confd emptyDir volume (deployment.yaml, volumes section), an init-redis-session-ini init container that mounts that volume at /usr/local/etc/php/conf.d and touches redis-session.ini into it, using the Nextcloud image so the file is owned by the right UID, and a subPath mount of that file into the nextcloud-cron sidecar.

The one container that actually needs it - the main nextcloud container, whose entrypoint writes the file - was never given the mount. This adds it, mirroring the existing block in the cron sidecar.

## Benefits

Fixes /entrypoint.sh: cannot create /usr/local/etc/php/conf.d/redis-session.ini: Permission denied when running the main container as non-root (e.g. nextcloud.securityContext.runAsUser: 33) with Redis enabled. /usr/local/etc/php/conf.d inside the image is root-owned, so the entrypoint's "Configuring Redis as session handler" step fails for any non-root user; the emptyDir the chart already provisions makes the single file writable without shadowing the rest of conf.d.

Users currently have to work around this in their own values with:

```
nextcloud:
  extraVolumeMounts:
    - name: php-confd
      mountPath: "/usr/local/etc/php/conf.d/redis-session.ini"
      subPath: redis-session.ini
```
which reaches into a chart-internal volume name. That workaround becomes unnecessary.

## Possible drawbacks

Anyone using the extraVolumeMounts workaround above will get a duplicate mountPath on the nextcloud container, which the API server rejects, until they remove it from their values. Worth a release note.
The mount stays gated on .Values.redis.enabled, consistent with the volume and the init container. externalRedis.enabled=true sets REDIS_HOST too, so it hits the same entrypoint path and the same permission error, but the volume and init container don't exist in that case - fixing external Redis means widening all four conditions and is left out of this PR deliberately.

## Applicable issues

relates to #187 - the original fix for this error; it covered the cron sidecar but not the main container
relates to #814 - same error reported on RKE2 with a non-root securityContext (closed without a chart-side fix)

## Additional information

Verified with helm template on the local chart:

redis.enabled=true, cronjob.enabled=true, nextcloud.securityContext.runAsUser=33 → the mount renders exactly once in the nextcloud container and once in nextcloud-cron; no duplicate mountPath in any container
redis.enabled=false → no php-confd references anywhere in the rendered Deployment
helm lint passes
No new or changed values, so no README changes are needed.

## Checklist 


edvinas31:
@promasu tagging you, but is there anyone able to look at this?

PR #881, opened 2026-08-24
feat(nextcloud): allow suspending the cron CronJob via values

marekjagielski:
## Description of the change

Renders `spec.suspend` on the cron CronJob unconditionally (defaulting to `false`), and exposes it as `cronjob.cronjob.suspend`.

## Benefits

The template currently renders no `spec.suspend` at all, so the field has **no desired value in the manifest**. Anything that reconciles the chart's output — ArgoCD, Flux, `helm diff` — has nothing to compare against, and under server-side apply a `kubectl patch ... suspend=true` takes ownership of the field and **survives every subsequent sync**, while the release keeps reporting as in-sync.

That is not hypothetical. We had `nextcloud-cron` suspended by hand during a debugging session and left that way for days. ArgoCD reconciled that very object repeatedly across the period and left the suspension in place — correctly, since it did not own the field. In that window Nextcloud ran no file scans, no trashbin/version expiry, no notification delivery, no federated sync and no app-update checks, and nothing surfaced it. `managedFields` showed `spec.suspend` owned by `kubectl-patch`, never by `argocd-controller`.

Rendering the field always puts it back under the chart's ownership, so a manual patch is reverted on the next sync instead of silently persisting. It also makes suspending the job a first-class values option rather than something you reach around the chart to do.

## Possible drawbacks

None that I can find. `false` is already the CronJob API default, so behaviour is unchanged for every existing release — the only difference is that a previously-omitted field now renders as an explicit `false`. For users who *deliberately* suspend the job out-of-band today, that manual patch will now be reverted on the next chart apply; setting `cronjob.cronjob.suspend: true` is the supported replacement, and arguably the point of the change.

## Applicable issues

None open that I could find — happy to file one first if you'd prefer that order.

## Additional information

Verified against `charts/nextcloud`:

- `helm lint .` — 1 chart linted, 0 failed
- `helm template .` with `cronjob.enabled=true cronjob.type=cronjob` → `suspend: false`
- same with `--set cronjob.cronjob.suspend=true` → `suspend: true`
- `cronjob.type=sidecar` renders no CronJob, unaffected
- documented in `values.yaml` and the README values table
- `Chart.yaml` bumped 9.2.6 → 9.3.0 (minor: adds a value, no breaking change)

## Checklist


charts/nextcloud/templates/cronjob.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
   {{- end }}
 spec:
   schedule: {{ .schedule | quote }}
+  suspend: {{ .suspend | default false }}
```
wrenix: The value file is already the place for default value

wrenix (approved):

charts/nextcloud/templates/cronjob.yaml
marekjagielski: Agreed — dropped in a78e2d8.

marekjagielski:
Friendly ping — the Lint and Test Charts run for a78e2d8 is sitting in `action_required` since 25 Aug and needs a maintainer to click *Approve and run*. The last commit only drops a redundant `| default false` from the template; rendering is unchanged. Happy to squash if that helps.

